PR #2, opened 2015-06-17
don't mask pixel channels for void-extent blocks in the sRGB mode

nanley:
According to section 3.1 (Decode Procedure) of the specification,
a color read from a void-extent block must be immediately returned.
Pixel channels are only masked when returning an interpolated color
from a non-void-extent block in the sRGB mode.

### DIFF
--- a/Source/astc_decompress_symbolic.cpp
+++ b/Source/astc_decompress_symbolic.cpp
@@ -132,13 +132,10 @@ void decompress_symbolic_block(astc_decode_mode decode_mode,
 
 		if (scb->block_mode == -2)
 		{
-			// For sRGB decoding, we should return only the top 8 bits.
-			int mask = (decode_mode == DECODE_LDR_SRGB) ? 0xFF00 : 0xFFFF;
-
-			red = sf16_to_float(unorm16_to_sf16(scb->constant_color[0] & mask));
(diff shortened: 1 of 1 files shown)


lordal01:
Hi Nanley,

thank you for submitting this, we are currently evaluating it and we will take action soon. 

Thanks,
Lorenzo

achary01:
Hi Nanley,
We have reviewed your patch proposal and we decided not to integrate it to the master branch.

Our view is:
The “immediately return single color” in spec as in 3.1. Decode Procedure chapter stands for returning the color from a void extent block without any interpolation operation as required for other block types. That’s the optimization gain. It should however not to be interpreted as doing literally no operation on it before the return.
Masking to upper 8 bits only is a correct operation in a context of sRGB encoded color values to keep 8-bit resolution as pointed in chapter 2.5. Table 4 (for LDR endpoint decoding precision).

If you have a particular use case where you change is vital for getting expected decoding result and where current reference implementation is not able to provide, please share it with us so we can investigate it further down.

Regards,
Adam

nanley:
Thank you for reviewing my patch proposal. I am currently getting unexpected output using the sRGB mode of the decoder. It seems that the issue lies in the fact that after masking out the bottom 8 bits, the pixel channels are treated as UNORM16s instead of UNORM8s. This brings in inaccuracies for input values above 128. I will send another patch for consideration.

Thanks,
Nanley

PR #3, opened 2015-06-21
use the top 8-bits more accurately for void-extent blocks in sRGB mode

nanley:
Section 2.5. Table 4, mentions that an 8-bit resolution is allowed for sRGB
mode. After masking the bottom 8 bits in the pixel channels of void-extent
blocks in sRGB mode, the resulting value is treated as a UNORM16. This causes
all input values above 128 (in the upper 8 bits) to return an output that
is 1 less than expected. For example, 129 in the top 8 bits returns 128.
This patch uses the top 8 bits to generate a float that will give the
expected output when writing out to disk.

### DIFF
--- a/Source/astc_decompress_symbolic.cpp
+++ b/Source/astc_decompress_symbolic.cpp
@@ -133,12 +133,18 @@ void decompress_symbolic_block(astc_decode_mode decode_mode,
 		if (scb->block_mode == -2)
 		{
 			// For sRGB decoding, we should return only the top 8 bits.
-			int mask = (decode_mode == DECODE_LDR_SRGB) ? 0xFF00 : 0xFFFF;
-
-			red = sf16_to_float(unorm16_to_sf16(scb->constant_color[0] & mask));
-			green = sf16_to_float(unorm16_to_sf16(scb->constant_color[1] & mask));
-			blue = sf16_to_float(unorm16_to_sf16(scb->constant_color[2] & mask));
-			alpha = sf16_to_float(unorm16_to_sf16(scb->constant_color[3] & mask));
+			if (decode_mode == DECODE_LDR_SRGB)
+			{
+				red   = (scb->constant_color[0] >> 8) / 255.0f;
+				green = (scb->constant_color[1] >> 8) / 255.0f;
+				blue  = (scb->constant_color[2] >> 8) / 255.0f;
+				alpha = (scb->constant_color[3] >> 8) / 255.0f;
+			} else {
+				red = sf16_to_float(unorm16_to_sf16(scb->constant_color[0]));
+				green = sf16_to_float(unorm16_to_sf16(scb->constant_color[1]));
+				blue = sf16_to_float(unorm16_to_sf16(scb->constant_color[2]));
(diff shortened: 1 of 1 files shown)


achary01:
Hi Nanley,
Thanks for input, we'll back to that then,
cheers,
Adam

NikunjPatel-ARM:
Could you give us a test case and the command you run so we can try and reproduce this issue on our end?

nanley:
I attempted to create a simple test case with a 4x4 image. I still get unexpected values, but I can't reason the cause (as I could in my current usage scenario). Although the patch fixes my current case, it actually produces worse results for the test case. Perhaps the fix isn't as simple as I had hoped. I can describe the issue I am currently encountering with sRGB decoding however.

The following are the steps I have running in a script [1] in which I am finding unexpected values being generated by the decoder:

* Use the image from [2] and rename it to waffles.jpg.

* Run the following commands: 
astcenc -cs waffles.jpg waffles.astc 4x4 -veryfast
astcenc -ds waffles.astc waffles.ktx

* Force OpenGL do an SRGB conversion by inputting the following values into the KTX header:
glType: GL_UNSIGNED_BYTE
glTypeSize: 1
glFormat: GL_RGB
glInternalFormat: GL_SRGB8
glBaseInternalFormat: GL_RGB

* Compare the pixels of two textures [3] using piglit. The first texture is the ASTC file decompressed by a HW decoder, the other is the KTX generated from the reference codec. The 4x4 void-extent block located at (x=288, y=0), where 0 is the bottom of the image, shows an unexpected output from the reference codec. The input at this block is 0xFDFDFD. The output value for each channel should be FD/FF = 0.992157 but instead it is 0.980392.

1. http://cgit.freedesktop.org/~nchery/ktx_utils/
2. http://people.freedesktop.org/~chadversary/permalink/2012-07-09/1574cff2-d091-4421-a3cf-b56c7943d060.jpg
3. http://cgit.freedesktop.org/~nchery/piglit/commit/?h=astc

NikunjPatel-ARM:
From what I can see, it looks like the problem is on your end. You appear to be doing linear reads from the compressed ASTC texture (getting you the unchanged 0xFDFDFD), but sRGB reads from the decompressed KTX.

The raw value of the decompressed KTX at the block you specified is also 0xFDFDFD (verified by converting the KTX using the Mali Texture Compression Tool to PNG and then inspecting using MS paint).

Gamma 2.2 (near enough sRGB) read of 0xFD (253):

FLOOR(POW((253 / 255), 2.2) * 255) = 250

250/255 = 0.980392

As far as I can tell, the software decoder is returning the exact same results as your hardware one.

nanley:
Thanks for the feedback and sorry for the late reply. The sRGB/linear read math makes sense. If I am doing linear reads from void-extent blocks, why do you think my patch gets rid of this mismatch? I don't think I completely understand all the steps involved with block decompression in this codec.

nanley:
I revisited my calculations and noticed that masking does not cause an off-by-one error as I originally described. Sorry for the trouble and thanks for looking into this.

PR #5, opened 2015-09-15
Fix possible out-of-bounds array access

hkr:
Changed implementation of four_partitions_find_best_combination_for_bitcount to match two_partitions_find_best_combination_for_bitcount and three_partitions_find_best_combination_for_bitcount.
We hit that case in Unity.

### DIFF
--- a/Source/astc_pick_best_endpoint_format.cpp
+++ b/Source/astc_pick_best_endpoint_format.cpp
@@ -695,7 +695,7 @@ static void four_partitions_find_best_combination_for_bitcount(float combined_be
 															   int bits_available, int *best_quantization_level, int *best_quantization_level_mod, int *best_formats, float *error_of_best_combination)
 {
 	int i;
-	int best_integer_count = -4;
+	int best_integer_count = 0;
 	float best_integer_count_error = 1e20f;
 	int integer_count;
 


NikunjPatel-ARM:
Hi, could you give us a reproducer for the OOB array access? I've got a feeling this pull request might be working around the bug rather than fixing it (namely that we're not initialising best_integer_count_error to a high enough number, causing best_integer_count to never get initialised).

It does seem kind of weird that best_integer_count gets initialised to -4 though.

hkr:
I don't have a simple repro application, I have to extract that from our codebase.
This might take a while.
It could be that we only hit that code path because of way we use the ASTC encoder.

hkr:
Would it help for me to find out if we hit the 'break' in line 707 or if the condition in line 709 is always false?

NikunjPatel-ARM:
If you could that would be great, thanks.

hkr:
We are accessing `quantization_mode_table[integer_count][bits_available]` (line 705) with `integer_count == 4` and `bits_available == 7`, get -1 and then hit the 'break'.

NikunjPatel-ARM:
From my quick investigation it does appear that your patch is a valid fix for the issue.

You have the dubious honour of being our first accepted pull request, meaning we've got some legal stuff to deal with before we can merge in the fix. We'll try to get back to you shortly.

PR #6, opened 2016-06-05
Vs2013

Kirpich30000:
Original code causes "invalidation" of a wrong partition candidate. Thereof in most cases partition_indices_2planes[0] and partition_indices_2planes[1] would be the same (because on the second iteration separate_errors[1st_best_partition] would still be the best). So effectively only one 2planes partition set would be tested.
 
This change provides minor quality improvement

### DIFF
--- a/Source/astc_codec_internals.h
+++ b/Source/astc_codec_internals.h
@@ -703,6 +703,28 @@ struct encoding_choice_errors
 	int can_blue_contract;
 };
 
+// buffers used to store intermediate data in compress_symbolic_block_fixed_partition_*()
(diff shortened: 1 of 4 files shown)


Kirpich30000:
Last change reduces compression time by ~20-30% on my test system (Intel Core i5 760)

Kirpich30000:
Add another change with a bug fix

lordal01:
Thank you, we are reviewing these.

NikunjPatel-ARM:
All the fixes look good from here.

Thanks!

PR #23, opened 2018-06-19
Update stb_image to 2.19

paraboul:
Newer version of stb image brings various improvement (speed, progressive jpeg support, 1/2/4-bit PNG support)


paraboul:
Note: I only updated the linux64 binary (no simple way to cross compile)

solidpixel:
I'm going to reject this one simply because it would be more useful to update with the latest version (2.22) because it includes some security fixes.

PR #47, opened 2019-07-03
Update documentation

solidpixel:
Added a new `Encoding.md` document to provide guidelines about how to actually drive the compressor command line options, in particular looking at format equivalency with other compression formats.

Moved build instructions to a new `Building.md` document, and added instructions for command line builds for Visual Studio.

Converted the license into `LICENSE.md` markdown, to make it a bit more readable in the browser (and no less readable in text format).

### DIFF
--- a/Docs/Encoding.md
+++ b/Docs/Encoding.md
@@ -151,7 +151,7 @@ Z component of the normal can be reconstructed in shader code based on the
 knowledge that the vector is unit length.
 
 To encode this we therefore want to store two input channels and should
-therefore using `rrrg` coding swizzle, and the `.ga` sampling swizzle. The
+therefore use the `rrrg` coding swizzle, and the `.ga` sampling swizzle. The
 OpenGL ES shader code for reconstruction of the Z value is:
 
     vec3 normal;
--- a/Docs/Encoding.md
+++ b/Docs/Encoding.md
@@ -1,0 +1,193 @@
+# Effective ASTC Encoding
+
+Most traditional texture compression schemes encode a single color format at
+single bitrate, so there are relatively few configuration options available to
+content creators beyond selecting which compressed format to use.
+
+ASTC on the other hand is an extremely flexible container format which can
+compress multiple color formats at multiple bit rates. Inevitably this
+flexibility gives rise to questions about how to best use ASTC to encode a
+specific color format, or what the equivalent settings are to get a close
+match to an existing legacy compression format.
+
+This page aims to give some guidelines, but note that they are only guidelines
+and are not exhaustive so please deviate from them as needed.
+
+## Traditional format reference
+
+The most commonly used traditional compressed formats, their color format, and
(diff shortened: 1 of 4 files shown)


Docs/Encoding.md
```diff
@@ -0,0 +1,193 @@
...
+controlled swizzles are not available in WebGL.
+
+# Other Considerations
+
+This section outlines some of the other things to consider when encoding
+textures using ASTC.
+
+## Encoding non-correlated channels
+
+ASTC can optionally assign any single color channel to a separate
```
mariusbjorge: Revise sentence

Docs/Encoding.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Encoding normal maps
+
+The best way to store normal maps using ASTC is similar to the scheme used by
+BC5; store the X and Y components of a unit-length tangent space normal. The
+Z component of the normal can be reconstructed in shader code based on the
+knowledge that the vector is unit length.
+
+To encode this we therefore want to store two input channels and should
+therefore using `rrrg` coding swizzle, and the `.ga` sampling swizzle. The
```
mariusbjorge: using -> use

README.md
```diff
@@ -137,6 +124,19 @@ There are many other command line options for tuning the encoder parameters
 which can be used to fine tune the compression algorithm. See the command line
 help message for more details.
 
+# Documentation
+
+The [Effective ASTC Encoding](./Docs/Encoding.md) page looks at some of the
+guidelines that should be followed when compressing data using `astcenc`.
+It covers:
+
+* How to efficiently encoding data with fewer than 4 channels.
```
mariusbjorge: encoding -> encode

Docs/Encoding.md
solidpixel: Can you clarify the problem here? It reads OK to me, but that's possibly just because I wrote it ...

Thanks!

solidpixel: Original text:

> ASTC can optionally assign any single color channel to a separate non-correlated partition to the other three channels, chosen on a block-by-block basis. In the general case there is no need for non-correlated data to be in a specific channel in the data set; the compressor will assign the non-correlated channel on demand if it is needed.

Attempted reword:

> Most other texture compression formats have a static channel assignment in terms of the expected data correlation. For example, ETC2+EAC assumes that RGB are always correlated and that alpha is non-correlated. ASTC can automatically encode data as either fully correlated across all 4 channels, or with any one channel assigned to a separate non correlated partition to the other three. The non-corellated channel can be changed on a block-by-block basis, so the compressor can dynamically adjust the coding based on the data present in the image. This means that there is no need for non-correlated data to be stored in a specific channel in the input image.

mariusbjorge: It was this part "to a separate non-correlated partition to the other three channels" - the reword looks much better.

PR #72, opened 2019-11-11
Remove variable shadows to pass GCC analysis

### LINKED ISSUE #70: fix variables shadowing
I'm getting compilation -Werror=shadow errors on GCC 7.4.0

solidpixel:
This fixes one bug in the code caused by shadowing where quant_mod
wasn't getting updated in the outer scope. Minor IQ differences
expected, on average it's better quality but really small differences
expected based on testing here (we will test more values for weight
perturbations).

Fixes #70

### DIFF
--- a/Source/astc_kmeans_partitioning.cpp
+++ b/Source/astc_kmeans_partitioning.cpp
@@ -110,11 +110,11 @@ void kpp_initialize(int xdim, int ydim, int zdim, int partition_count, const ima
 	// finally, gather up the results.
 	for (i = 0; i < partition_count; i++)
 	{
-		int sample = cluster_center_samples[i];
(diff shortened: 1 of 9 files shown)


Source/astc_kmeans_partitioning.cpp
```diff
@@ -142,7 +141,7 @@ void basic_kmeans_assign_pass(int xdim, int ydim, int zdim, int partition_count,
 							  blk->work_data[4 * i + 1],
 							  blk->work_data[4 * i + 2],
 							  blk->work_data[4 * i + 3]);
-		float4 diff = color - center_color;
+		float4 diff = color - cluster_centers[0];
```
renato-grottesi: is this intentional?

Issue #26, opened 2018-08-11
texture is y-flipped


face2d:
texture is y-flipped when load an ASTC-encoded (.astc) file, the astc data write to a texture's memory directly. But when I try to load ktx or pvr format file using astc-encoded, texture's content is normally. Is this a bug or a implicit rule ?

solidpixel:
It's expected behavior.

The original reference encoder was written against the Direct3D convention, for which the texture coordinate (0,0) is the top left corner of the texture.

The KTX wrapper is a Khronos standard, and for Khronos specifications such as OpenGL and OpenGL ES the texture coordinate (0,0) is in the bottom left corner of the texture.

Sorry - the industry just isn't self consistent here across APIs, so any single coordinate system is always wrong for someone.

PR #81, opened 2020-01-15
Added -flip to support OpenGL

Micha0:
As I was having the same issue as #26, I decided to fix it. Hopefully this inspires you to have this feature. For me the "DirectX argument" doesn't really fly as most desktop hardware doesn't support the ASTC format while a lot of new mobile phones do, where OpenGL is very popular. Since Vulkan and Metal do have the same texture coordinate system as DirectX, the default is kept unflipped.


solidpixel:
Happy for this to be merged to master; also tagged for 2.0 so we remember to pull this change across to the 2.0 branch.

Micha0:
Sounds fair @solidpixel, thank you. As this is now a feature request for 2.0, I guess you don't need documentation bit? (I'm not sure how much the 2.0 codebase is different from 1.x)

solidpixel:
No, thats fine - I can add the documentation when we do the 2.0 flavored version of this. I'll reject this PR - but I've raised an issue to track this for for the 2.0 branch - see  #82. Will try to get around to this soon, so you're not waiting for it for a long time (although beware the current 2.0 branch is very much an early alpha).

solidpixel:
FYI, `-yflip` now on master.

PR #88, opened 2020-02-17
CI: Add AppVeyor configuration

EwoutH:
Adds an AppVeyor configuration that builds `astcenc` with Visual Studio using the `astcenc.sln`.

AppVeyor can be [enabled](https://github.com/marketplace/appveyor) on the GitHub Marketplace.


solidpixel:
Thanks for the PR. This seems like a sensible thing to do, but I'll need to check with our commercial team here that we are happy to sign up with the AppVeyor T&C's.

solidpixel:
Rejecting.

PR #91, opened 2020-02-26
Memory leak fix and some cleanups

sugoi1:
- Fixed memory leak related to decimation_tables being
  allocated and never deallocated
- Reduced size of decimation_tables array to remove
  unused last element
- Made hdr_rgb_hdr_alpha_unpack3() static
- Removed spurious semicolons in vtypeX classes


solidpixel:
Thanks for the PR, much appreciated.

PR #113, opened 2020-06-08
Split out file loading from astc_main

solidpixel:
Pre-work for migrating to the new API, making astc_main less monolithic.

### DIFF
--- a/Source/astcenccli_image_load_store.cpp
+++ b/Source/astcenccli_image_load_store.cpp
@@ -1456,7 +1456,7 @@ astc_codec_image* load_dds_uncompressed_image(
 
 	delete[] buf;
 	fill_image_padding_area(astc_img);
-	is_hdr = bitness;
+	is_hdr = bitness == 16;
 	num_components = components;
 	return astc_img;
 }
--- a/Source/astcenccli_image_load_store.cpp
+++ b/Source/astcenccli_image_load_store.cpp
@@ -52,8 +52,9 @@ Image load and store through the stb_iamge and tinyexr libraries
 static astc_codec_image* load_image_with_tinyexr(
 	const char* filename,
 	int padding,
-	int y_flip,
-	int* result
+	bool y_flip,
+	bool& is_hdr,
+	int& num_components
 ) {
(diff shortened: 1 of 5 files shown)


Source/astcenccli_toplevel.cpp
```diff
@@ -399,6 +400,165 @@ static void test_inappropriate_cpu_extensions()
...
+}
+
+/**
+ * @brief Load a non-astc image file from memory.
+ *
+ * @param filename           The file to load, or a pattern for array loads.
+ * @param dim_z              The number of slices to load.
+ * @param padding            The number of texels of padding.
+ * @param y_flip             Should this image be Y flipped?
+ * @param linearize_srgb     Should this image be Y flipped?
```
ghufranjavaid, line 442: Should this image converted from linear to sRGB?

Source/astcenccli_toplevel.cpp
```diff
@@ -399,6 +400,165 @@ static void test_inappropriate_cpu_extensions()
...
+			fill_image_padding_area(image);
+		}
+
+		for (auto &i : slices)
+		{
+			free_image(i);
+		}
+	}
+
+	img_num_components = load_result & 7;
```
ghufranjavaid, line 556: Why are we multiplying it with 7? would be a good idea to have it as a const with a comment

solidpixel: The original code returns a packed int (bit[2..0] = num components, bit[3]  = 16 bit if set). Seems like needless optimization, so made that long-hand in the next commit.

Source/astcenccli_image_load_store.cpp
```diff
@@ -1468,7 +1456,8 @@ astc_codec_image* load_dds_uncompressed_image(
 
 	delete[] buf;
 	fill_image_padding_area(astc_img);
-	*result = components + (bitness == 16 ? 0x80 : 0);
+	is_hdr = bitness;
```
solidpixel, line 1459: Bug. Needs bitness == 16.

PR #119, opened 2020-06-10
Migrate CLI to use astcenc_image everywhere

solidpixel:


Source/astcenccli_error_metrics.cpp
```diff
@@ -139,31 +139,31 @@ void compute_error_metrics(
 	kahan_accum4 log_errorsum;
 	kahan_accum4 mpsnr_errorsum;
 
-	int xsize = MIN(img1->xsize, img2->xsize);
-	int ysize = MIN(img1->ysize, img2->ysize);
-	int zsize = MIN(img1->zsize, img2->zsize);
+	int xsize = MIN(img1->dim_x, img2->dim_y);
```
solidpixel: This is fixed in a later patch.

PR #147, opened 2020-09-09
Optimize trials of using 2/3 partitions and two planes of weights

solidpixel:
In the original code running trials for 2 or 3 partitions using two planes of weights, two different color components were tested as candidates for the second weight plane. The second component is rarely a good candidate, so the IQ benefits of this are tiny given the processing cost. 

For the Kodak test set with `-medium` compression we get an average of 8% performance improvement, for negligible quality loss (< 0.01 dB). 

```
Coding time:    Mean: +1.08x    Std: 0.03x
Image quality:  Mean: -0.00 dB  Std: 0.00 dB
```

There is a larger quality impact for `-through` and `-exhaustive` - although it is still very small (0.03 dB) - in particular for normal maps. However, in reality no developers are really using those search modes in production as they are far too slow so this seems acceptable.

### DIFF
--- a/Docs/Analysis.md
+++ b/Docs/Analysis.md
@@ -1,10 +1,10 @@
 # astcenc 2.1 optimization log
 
-This page is a working document explaining the my current understanding of the
+This page is a working document explaining the current understanding of the
 performance and quality impact of some of the compression trials, refinement
 passes, and the heuristics which drive them. To make things easier to
 cross-reference, I've inserted some comments in the code to identify passes
-that I've analyzed.
+that have been analyzed.
 
 This information will hopefully allow as "tuned up" a future version of the
 codec. At the moment it's just research.
--- a/Docs/Analysis.md
+++ b/Docs/Analysis.md
@@ -1,0 +1,26 @@
+# astcenc 2.0 analysis
+
+This page is a working document explaining my current understanding of the
+performance and quality impact of some of the compression trials, refinement
+passes, and the heuristics which drive them. To make things easier to
+cross-reference, I've inserted some comments in the code to identify passes
+that I've analyzed.
+
+This information will hopefully allow as "tuned up" a future version of the
+codec. At the moment it's just research.
+
+## Configuration sensitivity
+
+The most important thing to know for this analysis is that the impact of each
+trial and refinement can vary dramatically depending on the codec
+configuration. Principally this means:
+
+* block size,
+* quality preset,
+* image color format.
+
+For example, removing the initial 1 partition 1 plane fast-early-out
+optimization actually improves the performance of 4x4 blocks in `-fast` and
+`-medium`. However, it makes 5x5 blocks marginally slower and the slow down
+increases as block size increases.
+
--- a/Docs/Analysis.md
+++ b/Docs/Analysis.md
@@ -1,6 +1,6 @@
-# astcenc 2.0 analysis
+# astcenc 2.1 optimization log
 
-This page is a working document explaining my current understanding of the
+This page is a working document explaining the my current understanding of the
 performance and quality impact of some of the compression trials, refinement
 passes, and the heuristics which drive them. To make things easier to
 cross-reference, I've inserted some comments in the code to identify passes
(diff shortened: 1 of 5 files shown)


solidpixel:
There is more tidying up possible for `find_best_partitionings()` - will do that in a separate PR. This one just makes the gross change to stop returning two results.

Docs/Analysis.md
```diff
@@ -0,0 +1,26 @@
+# astcenc 2.0 analysis
+
+This page is a working document explaining my current understanding of the
+performance and quality impact of some of the compression trials, refinement
+passes, and the heuristics which drive them. To make things easier to
+cross-reference, I've inserted some comments in the code to identify passes
+that I've analyzed.
+
+This information will hopefully allow as "tuned up" a future version of the
+codec. At the moment it's just research.
```
BananaEarwig, line 10: codec as possible. At ...

Docs/Analysis.md
```diff
@@ -1,6 +1,6 @@
-# astcenc 2.0 analysis
+# astcenc 2.1 optimization log
 
-This page is a working document explaining my current understanding of the
+This page is a working document explaining the my current understanding of the
```
BananaEarwig, line 3: remove my

Source/astcenc_compress_symbolic.cpp
```diff
@@ -1221,8 +1221,8 @@ float compress_symbolic_block(
 
 	imageblock *temp = &tmpbuf->temp;
 
-	float best_errorvals_in_modes[17];
-	for (int i = 0; i < 17; i++)
+	float best_errorvals_in_modes[13];
+	for (int i = 0; i < 13; i++)
```
BananaEarwig, line 1225: magic numbers (13)

Source/astcenc_compress_symbolic.cpp
```diff
@@ -1380,7 +1385,8 @@ float compress_symbolic_block(
 				}
 			}
 
-			best_errorvals_in_modes[4 * (partition_count - 2) + 5 + i] = best_errorval_in_mode;
+			// Modes 5, 6, 8, 9, 11, 12
+			best_errorvals_in_modes[3 * (partition_count - 2) + 5 + i] = best_errorval_in_mode;
```
BananaEarwig, line 1389: Would it make sense to bounds check this calculation?

Source/astcenc_compress_symbolic.cpp
```diff
@@ -1399,19 +1405,16 @@ float compress_symbolic_block(
 			break;
 		}
 
-		for (int i = 0; i < 2; i++)
+
+		if (lowest_correl <= ctx.config.tune_two_plane_early_out_limit)
```
BananaEarwig, line 1409: Switched the direction comparison on this - intentional?

Source/astcenc_find_best_partitioning.cpp
```diff
@@ -719,99 +718,93 @@ void find_best_partitionings(
...
 		}
-
-		best_partitions_samechroma[i] = partition_sequence[best_samechroma_partition];
-		samechroma_errors[best_samechroma_partition] = 1e30f;
-		uncorr_errors[best_samechroma_partition] = 1e30f;
 	}
 
-	for (int i = 0; i < 2 * candidates_to_return; i++)
+	*best_partitions_samechroma = partition_sequence[best_samechroma_partition];
+	samechroma_errors[best_samechroma_partition] = 1e30f;
```
BananaEarwig, line 749: magic number 1e30f

BananaEarwig (approved):
Looks good, minor comments

Source/astcenc_compress_symbolic.cpp
solidpixel: For now I will leave this as the "minimal" edit to the existing code. I want to refactor this shortly; few of the values are actually used.

Source/astcenc_compress_symbolic.cpp
solidpixel: For now I will leave this as the "minimal" edit to the existing code. I want to refactor this shortly; few of the values are actually used.

Source/astcenc_compress_symbolic.cpp
solidpixel: Yes - it flipped from being a "disable if higher than limit" to being an "enable if lower than the limit".

PR #148, opened 2020-09-09
Skip dual weight plane searches for luminance-only blocks

solidpixel:
If you are using the luminance CEM, you will never need a second weight plane as you are only storing a single color value in the endpoint. This change early-outs the second weight plane search, giving substantial performance improvements to these block types, with +18% performance for a texture compressed using `-medium` and +30% for a texture compressed using `-thorough`. 

There is no image quality impact for this optimization; the work was completely redundant in this case.


BananaEarwig (approved):
Looks good

PR #149, opened 2020-09-09
Avoid 1 partition 1 plane early-out for small blocks

solidpixel:
The optimization is actually actually a de-optimization for both performance and image quality for 4x4 and 5x4 blocks. This PR changes it to be conditional based on current block size.

### DIFF
--- a/Source/astcenc_compress_symbolic.cpp
+++ b/Source/astcenc_compress_symbolic.cpp
@@ -1231,9 +1231,13 @@ float compress_symbolic_block(
 
 	float mode_cutoff = ctx.config.tune_block_mode_limit / 100.0f;
 
-	// next, test mode #0. This mode uses 1 plane of weights and 1 partition.
-	// we test it twice, first with a modecutoff of 0, then with the specified mode-cutoff.
-	// This causes an early-out that speeds up encoding of "easy" content.
+	// Trial using 1 plane of weights and 1 partition.
+
+	// Most of the time we test it twice, first with a mode cutoff of 0 and
+	// then with the specified mode cutoff. This causes an early-out that
+	// speeds up encoding of easy blocks. However, this optimization is
+	// disabled for 4x4 and 5x4 blocks where it nearly always slows down the
+	// compresion and slightly reduces image quality.
 
 	float modecutoffs[2];
 	float errorval_mult[2] = { 2.5, 1 };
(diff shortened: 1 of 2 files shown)


Source/astcenc_compress_symbolic.cpp
```diff
@@ -1231,17 +1231,24 @@ float compress_symbolic_block(
...
 
-	// next, test mode #0. This mode uses 1 plane of weights and 1 partition.
-	// we test it twice, first with a modecutoff of 0, then with the specified mode-cutoff.
-	// This causes an early-out that speeds up encoding of "easy" content.
+	// Trial using 1 plane of weights and 1 partition.
+
+	// Most of the time we test it twice, first with a mode cutoff of 0 and
+	// then with the specified mode cutoff. This causes an early-out that
+	// speeds up encoding of easy blocks. However, this optimization is
+	// disabled for 4x4 and 5x4 blocks where it nearly always slows down the
```
BananaEarwig, line 1239: Is that really 5x4 or should it be 5x5?

Source/astcenc_compress_symbolic.cpp
```diff
@@ -1231,17 +1231,24 @@ float compress_symbolic_block(
...
 	float modecutoffs[2];
 	float errorval_mult[2] = { 2.5, 1 };
 	modecutoffs[0] = 0;
 	modecutoffs[1] = mode_cutoff;
 
 	float best_errorval_in_mode;
-	for (int i = 0; i < 2; i++)
+
+
+	int start_trial = bsd->texel_count < 25 ? 1 : 0;
```
BananaEarwig, line 1250: 25 could maybe be something like MINIMUM_BLOCK_SIZE?

BananaEarwig (approved):
Looks good, a couple of minor nits

Source/astcenc_compress_symbolic.cpp
solidpixel: Yes, should be 5x4 - ASTC has many non-square to get fine-grained bitrate steppings.

Source/astcenc_compress_symbolic.cpp
solidpixel: Yes, good idea. The current code has a lot of "magic constants" like this - we should start pulling them out into named settings.

BananaEarwig (approved):
LGTM

BananaEarwig (approved):
LGTM

BananaEarwig (approved):
LGTM

PR #150, opened 2020-09-10
Refactor find_best_partitionings()

solidpixel:
We now only ever use a single result per block type from this function, so we can refactor to track the single best result as we go along instead of storing all results in an array that gets scanned after-the-fact.


BananaEarwig (approved):
Looks good

MulattoKid:
Looks good to me.

PR #165, opened 2020-10-23
Add "manual SIMD" utility (astcenc_vecmathlib.h) and use it in compute_lowest_and_highest_weight

aras-p:
`compute_lowest_and_highest_weight` today has manual SSE4.2 codepath and a scalar code path. Instead of that, replace it with an N-wide codepath using the new N-wide math utility header. With N being 8 (AVX2), 4 (SSE2/SSE4.2) or 1 otherwise.

I haven't done NEON side of this, but should be Fairly Easy ™️ to do.

On my machine (2018 MacBookPro, Core i9 2.9GHz), testing on a 2048x2048 texture with 6x6 blocks, medium preset, coding time:

- SSE2: 6.80s -> 6.10s
- SSE4.2: 5.43s -> 5.39s
- AVX2: 5.29s -> 5.02s

The PSNR between SSE2 and SSE4.2 was subtly different before, now they match. AVX2 PSNR is a bit different from those, but I suspect that's due to already existing AVX2 code path elsewhere (`compute_ideal_quantized_weights_for_decimation_table`).

The vecmathlib utility would need some more functions operands in order to use it in other places, like maybe replacing AVX2 codepath inside `compute_ideal_quantized_weights_for_decimation_table`, or "pick 4 best looking modes" loop inside `determine_optimal_set_of_endpoint_formats_to_use`. I can look into those if the overall approach makes sense.

### DIFF
--- a/Source/astcenc_vecmathlib.h
+++ b/Source/astcenc_vecmathlib.h
@@ -34,11 +34,91 @@
 #define SIMD_INLINE __attribute__((unused, always_inline, nodebug)) inline
 #endif
 
-#if ASTCENC_SSE != 0 || ASTCENC_AVX != 0
+#if ASTCENC_AVX >= 2
+    #define ASTCENC_SIMD_ISA_AVX2
+#elif ASTCENC_SSE >= 20
     #define ASTCENC_SIMD_ISA_SSE
 #else
     #define ASTCENC_SIMD_ISA_SCALAR
 #endif
+
+
+// ----------------------------------------------------------------------------
+// AVX2 8-wide implementation
+
+#ifdef ASTCENC_SIMD_ISA_AVX2
+
+#define ASTCENC_SIMD_WIDTH 8
+
+struct vfloat
+{
+    SIMD_INLINE vfloat() {}
+    SIMD_INLINE explicit vfloat(const float *p) { m = _mm256_loadu_ps(p); }
+    SIMD_INLINE explicit vfloat(float v) { m = _mm256_set1_ps(v); }
+    SIMD_INLINE explicit vfloat(__m256 v) { m = v; }
+    static vfloat zero() { return vfloat(_mm256_setzero_ps()); }
+    __m256 m;
+};
+
+struct vint
+{
+    SIMD_INLINE vint() {}
+    SIMD_INLINE explicit vint(const int *p) { m = _mm256_loadu_si256((const __m256i*)p); }
+    SIMD_INLINE explicit vint(int v) { m = _mm256_set1_epi32(v); }
+    SIMD_INLINE explicit vint(__m256i v) { m = v; }
+    __m256i m;
+};
+
+SIMD_INLINE vfloat load1a(const float* p) { return vfloat(_mm256_broadcast_ss(p)); }
+SIMD_INLINE vfloat loada(const float* p) { return vfloat(_mm256_load_ps(p)); }
+
+SIMD_INLINE vfloat operator+ (vfloat a, vfloat b) { a.m = _mm256_add_ps(a.m, b.m); return a; }
+SIMD_INLINE vfloat operator- (vfloat a, vfloat b) { a.m = _mm256_sub_ps(a.m, b.m); return a; }
+SIMD_INLINE vfloat operator* (vfloat a, vfloat b) { a.m = _mm256_mul_ps(a.m, b.m); return a; }
+SIMD_INLINE vfloat operator==(vfloat a, vfloat b) { a.m = _mm256_cmp_ps(a.m, b.m, _CMP_EQ_OQ); return a; }
+SIMD_INLINE vfloat operator!=(vfloat a, vfloat b) { a.m = _mm256_cmp_ps(a.m, b.m, _CMP_NEQ_OQ); return a; }
+SIMD_INLINE vfloat operator< (vfloat a, vfloat b) { a.m = _mm256_cmp_ps(a.m, b.m, _CMP_LT_OQ); return a; }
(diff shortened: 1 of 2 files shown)


Source/astcenc_weight_align.cpp
```diff
@@ -63,11 +73,12 @@ static const float angular_steppings[44] = {
 	20.0f, 21.0f, 22.0f, 23.0f,
 	24.0f, 25.0f, 26.0f, 27.0f,
 	28.0f, 29.0f, 30.0f, 31.0f,
-	32.0f, 33.0f, 34.0f, 35.0f
+	32.0f, 33.0f, 34.0f, 35.0f,
+#if ANGULAR_STEPS >= 48
+    32.0f, 33.0f, 34.0f, 35.0f,
```
solidpixel, line 78: Duplicating this row ensures we have the same results on 4 and 8 wide, which is important for users who want stable output across instruction sets. Seems a shame from a performance point of view, as it's dummy work.

Seems worth a comment to explain why we do it this way though.

aras-p: Added a comment in a8c2e89

Source/astcenc_weight_align.cpp
```diff
@@ -45,11 +45,22 @@
...
+    #define ANGULAR_STEPS 44
+#elif ASTCENC_SIMD_WIDTH == 8
+    // AVX code path loops over these tables 8 elements at a time,
+    // so make sure to have their size a multiple of 8.
+    #define ANGULAR_STEPS 48
+#else
+    #error Unknown SIMD width
+#endif
+
+static const float angular_steppings[ANGULAR_STEPS] = {
```
solidpixel, line 63: This needs to be `alignas(ASTCENC_VECALIGN) static const ...`, to ensure aligned loads work.

Source/astcenc_weight_align.cpp
```diff
@@ -62,11 +73,15 @@ static const float angular_steppings[44] = {
 	20.0f, 21.0f, 22.0f, 23.0f,
 	24.0f, 25.0f, 26.0f, 27.0f,
 	28.0f, 29.0f, 30.0f, 31.0f,
-	32.0f, 33.0f, 34.0f, 35.0f
+	32.0f, 33.0f, 34.0f, 35.0f,
+#if ANGULAR_STEPS >= 48
+    // This is "redundant" and only used in more-than-4-wide
```
solidpixel, line 78: Can we extend the comment with something along the lines of "The same value is replicated so that SSE and AVX code paths will return the same results.".

Source/astcenc_vecmathlib.h
```diff
@@ -0,0 +1,309 @@
+// SPDX-License-Identifier: Apache-2.0
```
solidpixel, line 1: General comment - can you rebase. I fixed an unrelated bug which was tripping our HDR tests, and I'd like to include that fix in the tests for this one before I merge it.

Source/astcenc_weight_align.cpp
```diff
@@ -174,143 +189,64 @@ static void compute_lowest_and_highest_weight(
...
+			minidx = select(minidx, idxv, mask);
+			vfloat accum = cut_low_weight_err + wt - vfloat(2.0f) * dwt;
+			cut_low_weight_err = select(cut_low_weight_err, accum, intAsFloat(mask));
+			
+			// Reset tracker on max hit.
+			mask = idxv > maxidx;
+			maxidx = select(maxidx, idxv, mask);
+			cut_high_weight_err = select(cut_high_weight_err, vfloat::zero(), intAsFloat(mask));
+			
+			// Accuulate on max hit.
```
solidpixel, line 229: Typo in accumulate

Source/astcenc_weight_align.cpp
```diff
@@ -174,143 +189,64 @@ static void compute_lowest_and_highest_weight(
...
+			mask = idxv == maxidx;
+			accum = cut_high_weight_err + wt + vfloat(2.0f) * dwt;
+			cut_high_weight_err = select(cut_high_weight_err, accum, intAsFloat(mask));
 		}
-
+
```
solidpixel, line 234: Coding standard nit, please strip any trailing whitespace.

Source/astcenc_weight_align.cpp
```diff
@@ -62,11 +73,15 @@ static const float angular_steppings[44] = {
...
+    // This is "redundant" and only used in more-than-4-wide
+    // SIMD code paths, to make the steps table size
+    // be a multiple of SIMD width.
+    32.0f, 33.0f, 34.0f, 35.0f,
+#endif
 };
 
-#define ANGULAR_STEPS ((int)(sizeof(angular_steppings)/sizeof(angular_steppings[0])))
-
 static float stepsizes[ANGULAR_STEPS];
```
solidpixel, line 85: Need `alignas(ASTCENC_VECALIGN)` on these two too.

solidpixel (changes requested):

aras-p:
Rebased and addressed the comments

solidpixel (dismissed):

Source/astcenc_weight_align.cpp
```diff
@@ -62,12 +73,18 @@ static const float angular_steppings[44] = {
...
+    // return the same results.
+    32.0f, 33.0f, 34.0f, 35.0f,
+#endif
 };
 
-#define ANGULAR_STEPS ((int)(sizeof(angular_steppings)/sizeof(angular_steppings[0])))
-
-static float stepsizes[ANGULAR_STEPS];
+alignas(ASTCENC_VECALIGN) static float stepsizes[ANGULAR_STEPS];
 static float stepsizes_sqr[ANGULAR_STEPS];
```
solidpixel, line 88: Need one more alignas here too - it's tripping up on some clang builds form me.

solidpixel (approved):

PR #166, opened 2020-10-26
More SIMD optimizations (optimize determine_optimal_set_of_endpoint_formats_to_use, compute_ideal_quantized_weights_for_decimation_table)

aras-p:
- Change vecmathlib header to have a separate `vmask` data type that is result of comparison operations, and an argument for `select()`
- Expand vecmathlib with additional operations (gather, horizontal min, extracting lane value etc.)
- Change loop in `determine_optimal_set_of_endpoint_formats_to_use` to be SIMD; this loops 4 times over 2048 sized arrays and was showing up in the profiler for me.
- Change `compute_ideal_quantized_weights_for_decimation_table` hardcoded 4-wide AVX2 loop to use vecmathlib, and so now it uses SIMD also on SSEn.

In my tests, on a 2048x2048 texture with 6x6 block sizes, -medium preset:
- Mac (2018 MacBookPro, Core i9 2.9GHz):
    - SSE2: 6.01s -> 5.20s
    - SSE4.2: 5.61s -> 5.13s
    - AVX2: 5.17s -> 4.74s
- Windows (AMD ThreadRipper 1950X 3.4GHz, 16 threads):
    - SSE2: 3.03s -> 2.89s
    - SSE4.2: 2.74s -> 2.52s
    - AVX2: 2.63s -> 2.59s

### DIFF
--- a/Source/astcenc_ideal_endpoints_and_weights.cpp
+++ b/Source/astcenc_ideal_endpoints_and_weights.cpp
@@ -24,6 +24,7 @@
 #include <cassert>
 
 #include "astcenc_internal.h"
+#include "astcenc_vecmathlib.h"
 
 #ifdef DEBUG_CAPTURE_NAN
 	#ifndef _GNU_SOURCE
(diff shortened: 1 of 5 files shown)


Source/astcenc_pick_best_endpoint_format.cpp
```diff
@@ -968,6 +971,33 @@ void determine_optimal_set_of_endpoint_formats_to_use(
...
 			}
 		}
+#else
+		// find best mode, SIMD N-wide way
+		vint vbest_error_index(-1);
+		vfloat vbest_ep_error(1e30f);
+		vint lane_ids = vint::lane_id();
+		for (int j = 0; j < MAX_WEIGHT_MODES; j += ASTCENC_SIMD_WIDTH)
+		{
+			vfloat err = vfloat(&errors_of_best_combination[j]);
```
solidpixel, line 981: Need to add `alignas(...)` on declaration of `errors_of_best_combination`, and `best_quantization_levels`.

Source/astcenc_pick_best_endpoint_format.cpp
```diff
@@ -968,6 +971,33 @@ void determine_optimal_set_of_endpoint_formats_to_use(
 				best_error_index = j;
 			}
 		}
+#else
+		// find best mode, SIMD N-wide way
+		vint vbest_error_index(-1);
+		vfloat vbest_ep_error(1e30f);
+		vint lane_ids = vint::lane_id();
+		for (int j = 0; j < MAX_WEIGHT_MODES; j += ASTCENC_SIMD_WIDTH)
```
solidpixel, line 979: It should always be the case, but can we add an assert that MAX_WEIGHT_MODES is an exact multiple of SIMD_WIDTH, just in case we optimize the weight modes used later.

Source/astcenc_vecmathlib.h
```diff
@@ -50,72 +50,213 @@
...
 SIMD_INLINE vint min(vint a, vint b) { a.m = _mm256_min_epi32(a.m, b.m); return a; }
 SIMD_INLINE vint max(vint a, vint b) { a.m = _mm256_max_epi32(a.m, b.m); return a; }
 
+// Horizontal minimum - returns vector with all lanes
+// set to the minimum value of the input vector.
+SIMD_INLINE vfloat hmin(vfloat v)
+{
+	__m128 vlow  = _mm256_castps256_ps128(v.m);
+	__m128 vhigh = _mm256_extractf128_ps(v.m, 1);
+		   vlow  = _mm_min_ps(vlow, vhigh);
```
solidpixel, line 201: Indent issues? (possibly just github review formatting)

Source/astcenc_ideal_endpoints_and_weights.cpp
```diff
@@ -1115,48 +1116,39 @@ void compute_ideal_quantized_weights_for_decimation_table(
...
+	vfloat scaled_low_boundv(scaled_low_bound);
+	vfloat quant_level_m1v(quant_level_m1);
+	vfloat rscalev(rscale);
+	vfloat low_boundv(low_bound);
+	for (/*Vector loop */; i < clipped_weight_count; i += ASTCENC_SIMD_WIDTH)
 	{
-		__m128 ix = _mm_load_ps(&weight_set_in[i]);
-		ix = _mm_mul_ps(ix, scalev);
-		ix = _mm_sub_ps(ix, scaled_low_boundv);
+		vfloat ix = loada(&weight_set_in[i]) * scalev - scaled_low_boundv;
```
solidpixel, line 1129: The data storage for `weight_set_in` and friends are allocated as part of the `compress_fixed_partition_buffers` scratch buffer struct allocated for each thread. We allocate the structure with `alignas()`, but we don't tag the individual members so it's possible they could unalign in future if we change that structure. We should add alignas() to that struct for all of the individual members (at least the bottom 4 entries).

solidpixel (changes requested):

Source/astcenc_pick_best_endpoint_format.cpp
solidpixel: Unresolving. 

Still missing alignas on best_quantization_levels.

solidpixel (changes requested):
One remaining missing alignas, but other than that that , this one passed image testing. Helps -fast the most, in percentage terms the new optimization only gives minor benefit to -throughput runs as other parts of the codec are more significant.

aras-p:
d'oh, sorry! Fixed now.

What are the speedups you see btw? I see the most (as expected) on SSE2/SSE4.2, with older AMD Ryzen being almost no improvement on AVX2 (it's dual-pumped IIRC, so 8-wide is not really faster than 4-wide), but AVX2 on Intel CPU does see a speedup.

solidpixel:
> What are the speedups you see btw?

Roughly, with AVX2 on an Intel i5-i9600K:

- `-fast`: seeing 1-3% for HDR images and normal maps, 2-7% for LDR image
- `-medium`: seeing 1-2% for HDR images and normal maps, 1-4% for LDR image
- `-thorough`: seeing proportionally less - around 0-2% for LDR images

I've not historically tracked pure SSE2, so I'm expecting great speedups there if we replace all of the existing hand-coded paths with the new vector maths support (given there wasn't an intrinsics path there at all). It won't quite catch SSE 4.2, but it should be pretty close.

solidpixel:
Here's the full performance report for avx2: [pr166-report.txt](https://github.com/ARM-software/astc-encoder/files/5440955/pr166-report.txt)

aras-p:
> if we replace all of the existing hand-coded paths

That reminds me -- how many of hardcoded paths there are today, actually? I don't see many -- there's some paths that are writtten in `astcenc_mathlib.h` that are not active (guarded by `&& USE_SCALAR_SSE` which is zero). I only see `flt_rte()` and `flt_rd()` functions that use SSE (4.2).

solidpixel:
> how many of hardcoded paths there are today, actually?

There weren't many - just three or four actual "algorithms" where we did intrinsics for the whole code path. Grep says that it looks like you might have hit them all already =)

The `USE_SCALAR_SSE` paths were just for testing some algorithmic changes before vectorization - expect the performance sucks for real usage so it's off by default.

PR #167, opened 2020-10-27
simd: NEON code path in vecmathlib

aras-p:
On iPhone 11 Pro, compressing 2048x2048 image at 6x6 block size, -medium, coding time goes from 16.05s down to 10.61s. PSNR a tiny bit different (41.827408 dB -> 41.827412 dB) due to `flt2int_rtn()` not behaving the same as a proper "round to nearest, even when a tie".


solidpixel:
Rejecting, as superseded by PR #191. Thanks for the starting point =)

PR #168, opened 2020-10-27
simd: vectorize compute_angular_offsets and atan2

aras-p:
For the SIMD `atan2` I went with the same approximation as already is in `astc::atan2`. There's a commented out code path with a more exact approximation, for now went for identical results as before.

Also, AVX2 now actually returns identical result (PSNR at least) compared to SSE2/SSE4.2

On 2048x2048 image, 6x6 block size, -medium, coding time:
- Mac (2018 MBP, Core i9 2.9GHz):
    - SSE2: 5.20s -> 5.05s
    - SSE4.2: 5.13s -> 4.49s
    - AVX2: 4.74s -> 4.28s
- Windows (AMD ThreadRipper 1950X, 3.4GHz, 16 threads):
    - SSE2: 2.89s -> 2.76s
    - SSE4.2: 2.52s -> 2.39s
    - AVX2: 2.59s -> 2.52s

PSNR unchanged

This would "logically conflict" with NEON (#167) since that one would be missing some of the ops I've added here (`abs`, division etc.). I can rebase and fixup that one onto this if needed.

### DIFF
--- a/Source/astcenc_vecmathlib.h
+++ b/Source/astcenc_vecmathlib.h
@@ -542,6 +542,10 @@ SIMD_INLINE void print(vint a)
 
 #ifdef ASTCENC_SIMD_ISA_SCALAR
 
+#include <algorithm>
+#include <math.h>
+#include <string.h>
+
 #define ASTCENC_SIMD_WIDTH 1
 
 struct vfloat
@@ -596,7 +600,7 @@ SIMD_INLINE vfloat min(vfloat a, vfloat b) { a.m = a.m < b.m ? a.m : b.m; return
 SIMD_INLINE vfloat max(vfloat a, vfloat b) { a.m = a.m > b.m ? a.m : b.m; return a; }
 SIMD_INLINE vfloat saturate(vfloat a) { return vfloat(std::min(std::max(a.m,0.0f), 1.0f)); }
 
-SIMD_INLINE vfloat abs(vfloat x) { return vfloat(std::abs(a.m)); }
+SIMD_INLINE vfloat abs(vfloat x) { return vfloat(std::abs(x.m)); }
 
 SIMD_INLINE vfloat round(vfloat v)
 {
--- a/Source/astcenc_vecmathlib.h
+++ b/Source/astcenc_vecmathlib.h
@@ -130,6 +130,7 @@ SIMD_INLINE vfloat loada(const float* p) { return vfloat(_mm256_load_ps(p)); }
 SIMD_INLINE vfloat operator+ (vfloat a, vfloat b) { a.m = _mm256_add_ps(a.m, b.m); return a; }
 SIMD_INLINE vfloat operator- (vfloat a, vfloat b) { a.m = _mm256_sub_ps(a.m, b.m); return a; }
 SIMD_INLINE vfloat operator* (vfloat a, vfloat b) { a.m = _mm256_mul_ps(a.m, b.m); return a; }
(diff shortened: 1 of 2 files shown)


Source/astcenc_vecmathlib.h
```diff
@@ -130,6 +130,7 @@ SIMD_INLINE vfloat loada(const float* p) { return vfloat(_mm256_load_ps(p)); }
 SIMD_INLINE vfloat operator+ (vfloat a, vfloat b) { a.m = _mm256_add_ps(a.m, b.m); return a; }
 SIMD_INLINE vfloat operator- (vfloat a, vfloat b) { a.m = _mm256_sub_ps(a.m, b.m); return a; }
 SIMD_INLINE vfloat operator* (vfloat a, vfloat b) { a.m = _mm256_mul_ps(a.m, b.m); return a; }
+SIMD_INLINE vfloat operator/ (vfloat a, vfloat b) { a.m = _mm256_div_ps(a.m, b.m); return a; }
```
solidpixel, line 133: Not one for this PR, but if we're happy with a fast approximation it might be worth trying rcp and mul here, as div is quite a lot slower (13 cycles on Haswell for div, vs 2.5 cycles for rcp + mul).

solidpixel: Alternatively, providing a fastdiv function so it can be used selectively.

Source/astcenc_weight_align.cpp
```diff
@@ -135,15 +135,11 @@ static void compute_angular_offsets(
...
-
-	for (int i = 0; i < max_angular_steps; i++)
-	{
-		anglesum_x[i] = 0;
-		anglesum_y[i] = 0;
-	}
-
+	alignas(ASTCENC_VECALIGN) float anglesum_x[ANGULAR_STEPS];
+	alignas(ASTCENC_VECALIGN) float anglesum_y[ANGULAR_STEPS];
+	memset(anglesum_x, 0, max_angular_steps*sizeof(anglesum_x[0]));
```
solidpixel, line 140: As these are both arrays of known size you can just do `sizeof(anglesum_x)`, etc.

Source/astcenc_weight_align.cpp
```diff
@@ -156,21 +152,25 @@ static void compute_angular_offsets(
 		const float *sinptr = sin_table[isample];
 		const float *cosptr = cos_table[isample];
 
-		for (int j = 0; j < max_angular_steps; j++)
+		vfloat sample_weightv(sample_weight);
+		for (int j = 0; j < max_angular_steps; j += ASTCENC_SIMD_WIDTH) // arrays are multiple of SIMD width (ANGULAR_STEPS), safe to overshoot max
```
solidpixel, line 156: Can we add a static_assert for the ANGULAR_STEPS being a multiple of SIMD_WIDTH.

Source/astcenc_vecmathlib.h
aras-p: Maybe as an option, but precision of rcp is different between Intel & AMD CPUs, and I'd rather avoid that. We have tweaked another texture compression (ISPC one) to avoid this for exact this reason.

solidpixel: Didn't know that one - that's a pain.

Source/astcenc_weight_align.cpp
aras-p: that would memset the whole array, whereas only max_angular_steps portion is needed

Source/astcenc_weight_align.cpp
aras-p: Done in 86c1dd8

PR #169, opened 2020-10-29
Densely pack valid block weight modes (10-20% speedup)

aras-p:
Since there's only a small subset of 2048 ones that are valid. For 2D LDR block sizes, this is the valid mode count -- 4x4: 145, 5x5: 255, 6x6: 370, 8x8: 555, 10x10: 687, 12x12: 775. But the code was looping over all the possible 2048 ones, doing checks "is this valid?" or sentinel values to skip over them.

Coding time of 2048x2048 image:

- 6x6 medium: Mac 4.28s -> 3.81s, PC 2.38s -> 2.22s
- 6x6 fast: Mac 0.72s -> 0.57s, PC 0.64s -> 0.57s
- 4x4 medium: Mac 8.53s -> 6.47s, PC 3.92s -> 3.28s
- 4x4 fast: Mac 1.27s -> 0.88s, PC 0.95s -> 0.76s

(Mac being 2018 MBP, Core i9 2.9GHz; PC being AMD ThreadRipper 1950X 3.4GHz; both tested with AVX2 code path)

It's entirely possible I have messed something up, e.g. missed an index update somewhere. That said, PSNR on this one image I tested on is exactly the same before & after the change.

### DIFF
--- a/Source/astcenc_block_sizes2.cpp
+++ b/Source/astcenc_block_sizes2.cpp
@@ -659,12 +659,12 @@ static void construct_block_size_descriptor_2d(
 #endif
 
 	// then construct the list of block formats
+	int idx = 0;
 	for (int i = 0; i < 2048; i++)
 	{
 		int x_weights, y_weights;
 		int is_dual_plane;
 		int quantization_mode;
-		int fail = 0;
 		int permit_encode = 1;
 
 		if (decode_block_mode_2d(i, &x_weights, &y_weights, &is_dual_plane, &quantization_mode))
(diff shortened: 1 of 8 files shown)


Source/astcenc_block_sizes2.cpp
```diff
@@ -659,12 +659,12 @@ static void construct_block_size_descriptor_2d(
 #endif
 
 	// then construct the list of block formats
+	int idx = 0;
```
solidpixel, line 662: We get some name shadowing errors if compiled with `-Werror=shadow` (some of the code lower down also has an `idx` variable). Can we rename the new variable to `bm_idx`, or something like that; shadowing makes future code review harder.

Source/astcenc_block_sizes2.cpp
```diff
@@ -849,12 +841,12 @@ static void construct_block_size_descriptor_3d(
 	bsd->decimation_mode_count = decimation_mode_count;
 
 	// then construct the list of block formats
+	int idx = 0;
```
solidpixel, line 844: We get some name shadowing errors if compiled with `-Werror=shadow` (some of the code lower down also has an `idx` variable). Can we rename the new variable to `packed_idx`, given you use this elsewhere already for this.

solidpixel (changes requested):

solidpixel:
Other than the name shadowing problems, which trip our compiles, the tests pass OK for me - thrown LDR, HDR, normals, and 3D at it and PSNR is identical.

Source/astcenc_compress_symbolic.cpp
```diff
@@ -306,9 +309,10 @@ static void compress_symbolic_block_fixed_partition_1_plane(
 	int qwt_bitcounts[MAX_WEIGHT_MODES];
 	float qwt_errors[MAX_WEIGHT_MODES];
 
-	for (int i = 0; i < MAX_WEIGHT_MODES; i++)
+	for (int i = 0, ni = bsd->block_mode_packed_count; i < ni; ++i)
```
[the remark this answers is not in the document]
aras-p: I kinda got used to this pattern to "make sure" that the compiler does not decide to do a load of `bsd->block_mode_packed_count` on each iteration of the loop (which it may or might not realize is possible)

Source/astcenc_block_sizes2.cpp
aras-p: This and other comments addressed in 6b0dc2b

solidpixel (approved):

PR #171, opened 2020-10-30
Remove the scratch copy of the original data.

solidpixel:
The current code makes a copy of the original data (`imageblock.orig_data`), transferring into working buffers (`imageblock.data_[rgba]`) for each encoding trial. The working buffer is overwritten by each trial, so we have to repopulate the working buffer after each trial which is a significant (10-15%) overhead for the fast searches. In reality the output of the decompression is never really used - we only need to compute RMS error for the trial (which can be done inline) and throw the output away.

This patch removes the `orig_data` copy of the data, and uses only the `data_[rgba]` fields. Data encoding transforms at the start/end of the block are done in-place, and the data is no longer overwritten during each trial. Some parts of the code needed to be adapted to the fact that the "working" encoding of the data is not the original encoding.

PSNR is identical for LDR images, but has minor changes for HDR images. This is because we now compute derivatives on a converted form of the modified data, so we have an additional conversion round-trip which can add some rounding. On our test set this actually slightly improves PSNR, but I suspect this is just luck.


Source/Makefile
```diff
@@ -118,7 +118,7 @@ endif
 
 # Validate that the BUILD parameter is a supported value, and patch CXXFLAGS
 ifeq ($(BUILD),release)
-CXXFLAGS += -O3 -DNDEBUG -flto
+CXXFLAGS += -O3 -DNDEBUG -flto -g
```
solidpixel, line 121: Leak from the debug session; remove.

.gitignore
```diff
@@ -6,6 +6,7 @@
 *.user
 __pycache__
 Scratch
+SourceOrig
```
solidpixel, line 9: Leak from the debug session; remove.

Source/astcenc_compress_symbolic.cpp
```diff
@@ -1141,9 +1153,10 @@ static void prepare_block_statistics(
 	float nf_sum = 0.0f;
 	for (int i = 0; i < texels_per_block; i++)
 	{
-		float3 val = float3(blk->orig_data[4 * i],
-							blk->orig_data[4 * i + 1],
-							blk->orig_data[4 * i + 2]);
+		float3 val = float3(blk->data_r[i],
```
solidpixel, line 1156: Whitespace?

Source/astcenc_image.cpp
```diff
@@ -107,15 +107,19 @@ void imageblock_initialize_deriv(
...
-	const float *fptr = pb->orig_data;
 	for (int i = 0; i < pixelcount; i++)
 	{
 		// compute derivatives for RGB first
 		if (pb->rgb_lns[i])
 		{
-			float r = MAX(fptr[0], 6e-5f);
-			float g = MAX(fptr[1], 6e-5f);
-			float b = MAX(fptr[2], 6e-5f);
+            float3 fdata = float3(pb->data_r[i], pb->data_g[i], pb->data_b[i]);
```
solidpixel, line 115: Whitespace

PR #176, opened 2020-11-08
Correct block footprint in bit-rate table documentation

jeremyong:


jeremyong:
The bit-rate table in the documentation [here](https://github.com/ARM-software/astc-encoder/blob/master/Docs/FormatOverview.md#block-sizes) lists the 8x8 footprint twice with two different bit rates. I believe the 3.2 bits/texel rate corresponds to the 8x5 footprint but please correct this commit if that is incorrect.

Thanks

solidpixel (approved):

solidpixel:
Thanks for the PR. Merged.

PR #182, opened 2020-11-23
First pass at new API for 2D slice image

solidpixel:
This implementation changes the core codec API for images so they they are no longer 3D arrays (which require a pointer tree to get allocated and created). Instead images are passed as an array of 2D image slices, building up 3D images in layers. Each slice consists of tightly packed 4 channel data with no padding between rows. 

The common case usage of ASTC is 2D images, so in the common case the slice array consists of just a single slice, while leaving support for 3D as an option in the API when it's needed. In addition passing raw data slices loaded from other libraries is now possible, as there is no longer need for the pointer preamble. This can remove additional memory allocation and copy overheads

Slight performance improvement for `-fastest` is expected - but will be < 1%. The main improvement here is just improved ease of integration with data from other sources, and the ability to avoid two memory allocations for the same data. 

Note that this change is ONLY the API change; the command line wrapper is still creating copies of the data from stb_image and tiny_exr, which isn't really needed. This will get fixed in a later PR.

### DIFF
--- a/Docs/ChangeLog.md
+++ b/Docs/ChangeLog.md
@@ -24,7 +24,9 @@ recompile your client-side code using the updated `astcenc.h` header.
   * **API Change:** Images using region-based metrics no longer need to include
     padding; all input images should be tightly packed and `dim_pad` is removed
     from the `astcenc_image` structure.
-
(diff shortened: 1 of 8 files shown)


BananaEarwig (approved):
Looks good

PR #183, opened 2020-12-01
Add image preprocessing options

solidpixel:
Add some commonly requested image preprocessing options to the command line front-end. Note that these are entirely local to the front-end, they are not part of the core codec/library interface.

The `-pp-normalize` option is designed to work with the `-normalize` option, and forces normal vectors in a normal map to be unit length. This is useful when compressing source textures that use normal length to encode an NDF, as per the M. Toksvig paper, which is incompatible with ASTC's two channel encoding that relies on normals being unit length.

The `-pp-premultiply` option is used to scale RGB by the alpha value to minimize cross-channel color bleed caused by GPU post-multiply filtering/blending.


MulattoKid:
Looks good to me.

PR #186, opened 2020-12-14
Port builds to use CMake

solidpixel:
This PR changes to build system from separate Makefile (g++/Clang) and Solution files (MSVC) to a unified approach using CMake as a build system generator.

- [x] Updating Jenkins and testing on Arm CI build infrastructure
- [x] Updating tests to find binaries under test in CMake install directory
- [ ] Performance testing to compare against old Make builds
- [x] Updating Markdown for build instructions
- [x] Packaging using the same layout as we use for GitHub releases


solidpixel:
Marking as draft. Further work needed:

* Updating Jenkins and testing on Arm CI build infrastructure
* Updating tests to find binaries under test in CMake install directory
* Updating Markdown for build instructions
* Performance testing to compare against old Make builds
* Packaging using the same layout as we use for GitHub releases

CMake/sign-exe.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
bengaineyarm: I don't think we should be shipping this file publicly

solidpixel: Already gone in the latest commit. Will need to find a way to patch the checkout in Arm CI though - I do want the ability to sign the GitHub releases.

bengaineyarm: I think you can probably build the signing tools into some docker image or install them on the build node for windows and have jenkins invoke them directly

Source/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+        $<$<CXX_COMPILER_ID:GNU,Clang>:-Wshadow>)
+        $<$<CXX_COMPILER_ID:GNU,Clang>:-Wdouble-promotion>
+
+target_include_directories(astcenc-${VEC}
+    PRIVATE
+        ${CMAKE_CURRENT_BINARY_DIR})
+
+target_link_libraries(astcenc-${VEC}
+    PRIVATE
+        $<$<CXX_COMPILER_ID:GNU,Clang>:pthread>)
```
bengaineyarm: Probably better to use PLATFORM_ID:Linux,Darwin rather than using the compiler to determine whether or not to use pthreads.
Additionally, i think this is better as a compiler option rather than a library

PR #199, opened 2021-01-11
Use template clamp/min/max when possible

solidpixel:


solidpixel:
Bad PR - will reupload

PR #207, opened 2021-01-30
Allow astc-encoder to be built as a static library

Themaister:
Currently the CMake build does not support building the API as a static library.

I'm not sure what the right approach is here. I had to hack together static library support for Granite integration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,9 +16,9 @@
 #  ----------------------------------------------------------------------------
 
 # CMake configuration
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.10)
 cmake_policy(SET CMP0069 NEW)  # LTO support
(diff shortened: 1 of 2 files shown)


solidpixel:
Happy to take a look - ability to build as a static/dynamic library has been on my "todo" for a while.  Only caveat here is that the core codec interface isn't pure C (we use references) so it's only going to reliably work on the same toolset, etc (due to name mangling differences across compilers).

Themaister:
Static lib is fine for that, especially in CMake, since I do `target_link_libraries(engine PRIVATE astc-encoder)`.

solidpixel:
> On that note, is there any reason why an API that looks very much like C is actually a C++-only API?

Not really - given the core code is C++ I just defaulted to writing the API in C++.

solidpixel:
Merged the static library changes to `main` - tested locally on Linux and Windows, and will let the macOS build run overnight to check.

PR #209, opened 2021-02-03
Replace -a zero-weight blocks with constant color blocks

solidpixel:


Source/astcenc_entry.cpp
```diff
@@ -717,9 +717,48 @@ static void compress_image(
...
+				// Do we have any alpha values?
+				use_full_block = false;
+				for (int ay = start_y; ay < end_y; ay++)
+				{
+					for (int ax = start_x; ax < end_x; ax++)
+					{
+						float a_avg = ctx.input_alpha_averages[ay * dim_x + ax];
+						// SATs accumulate error, so don't test exactly zero
+						// a=0x1 for single pixel in 12x12 block = 2.7e-5 avg
+						if (a_avg > 2.6e-5f)
```
solidpixel, line 739: This isn't quite true. We need to watch for contribution from pixels outside the block, so the total area that should be considered is `block_size + 2 * alpha radius`.

PR #235, opened 2021-03-08
Support RGBM map compression

solidpixel:
This branch is an implementation of a set of format-aware heuristics for compressing RGBM-encoded textures (see issue #236 for details). Due to how ASTC compresses, there are some recommendations for the input RGBM encoding, arbitrary data with a naive RGBM encoding will likely compress badly. 

The user must pre-encode the RGBM data; the API/CLI expects to have the data passed in RGBM format. During encoding the user must keep the resulting M value above a minimum threshold (32 recommended), in order to avoid quantization during ASTC compression rounding M to zero and resulting in black/white pixels. This thresholding reduces the precision of dark values, but the result is still at least 8-bit precision for the LDR range, and the HDR range is not impacted by the clamping. This is not enforced, but you may get severe block artifacts without a clamp. A standalone sample utility for applying RGBM encoding and decoding to an uncompressed image, including support for thresholding, can be found in `./Utils/astc_rgbm_codec.cpp`.

To compress an RGBM-encoded image using the command line use the new `-rgbm <multiplier>` command line option, where the multiplier is the scale value used during reconstruction (`vec3 hdr_val = tex.rgb * tex.a * <multiplier>`). This is used to correctly assess candidate encodings.

To compress using the API create the config using the new `ASTCENC_FLG_MAP_RGBM` flag. This will configure the codec for an RGBM scale of 5 by default. If your RGBM scale is not 5, then you must change `config.rgbm_m_scale` to your scale factor and `config.cw_a_weight` to double your scale factor before creating the context. 

The decompress using the API or the command line no special action is required; the additional heuristics only impact the compression path.


solidpixel:
Random idea would be to offset encode the M so that the shader reconstruction injects the minimal base offset. E.g. something like:

```
a.rgb = a.rgb * (5.0 * (a.m + 0.05));
```

This guarantees the encoding can't have a zero M value (black is still possible - encoded through the RGB channels), although it still might have high differential pixels.

PR #239, opened 2021-03-15
Fixed uninitialized memory access

akb825:
Check block_mode_packed_index before other block mode properties, since they are never set if the block mode is invalid.

This is the code in `construct_block_size_descriptor_2d()` that causes the memory to be uninitialized:

```
// Skip modes that are invalid, too large, or not selected by heuristic
if (!valid || !selected || (x_weights > x_dim) || (y_weights > y_dim))
{
	bsd.block_mode_packed_index[i] = -1;
	continue;
}
```

### DIFF
--- a/Source/astcenc_block_sizes2.cpp
+++ b/Source/astcenc_block_sizes2.cpp
@@ -802,6 +802,11 @@ static void construct_block_size_descriptor_2d(
 			bsd.block_modes[packed_idx].percentile_hit = true;
 			bsd.decimation_modes[decimation_mode].percentile_hit = true;
 		}
+		else
+		{
+			bsd.block_modes[packed_idx].percentile_always = false;
+			bsd.block_modes[packed_idx].percentile_hit = false;
+		}
 #endif
 
 		bsd.block_modes[packed_idx].decimation_mode = decimation_mode;


akb825:
After applying this fix I get the following assert in astcenc_symbolic_physicial.cpp when using `ASTCENC_PRE_FASTEST`:

> assert(packed_index >= 0 && packed_index < bsd.block_mode_count);

My guess is that previously the uninitialized memory was causing invalidated blocks to be processed, but now that my fix prevents those blocks from being processed the error threshold from being met.

solidpixel:
Can you share a reproducer for the original uninitialized memory bug you were seeing? This patch might hide it, but I don't think it is a valid fix at the moment.

The most critical thing is that the loops you patched are iterating over the number of packed block modes. The `bsd.block_mode_count` tracker only includes the valid block modes that are kept after filtering has been applied.

Modes that are skipped by this check `if (!valid || !selected || (x_weights > x_dim) || (y_weights > y_dim))` never increment the packed index, so they are never even stored and are never iterated by the trackers you patched.

The bit that is wrong in this patch is this bit `bsd->block_mode_packed_index[i] < 0`. This is using the packed index to look up in the unpacked-to-packed index translation, which doesn't make any sense.

akb825:
I encountered this with my project [Cuttlefish](https://github.com/akb825/Cuttlefish) when I refactored my ASTC conversion code to use the public API and cache context objects. ([direct link to the file that calls into astenc](https://github.com/akb825/Cuttlefish/blob/master/lib/src/AstcConverter.cpp))

This showed up as a crash when attempting to convert a texture in debug mode. It showed up after making the refactors, even after a full rebuild, which I thought was due to the change in memory layout. I tried reverting my change right now to reproduce the crash to provide the location it was breaking at, but can't reproduce it at the moment, though if it's uninitialized memory it will depend on whatever the memory state happens to be at the time. IIRC it was somewhere inside of `compress_symbolic_block_fixed_partition_2_planes()`, and crashed when attempting to assign a `vtype4`, complaining that the vector was inaccessible.

I tried running through valgrind, and while it didn't reproduce the crash itself, it did report uninitialized memory reads at astcenc_weight_align.cpp line 373 (`if (bm.is_dual_plane || (only_always && !bm.percentile_always) || !bm.percentile_hit)`) and   astcenc_compress_symbolic.cpp line 328 (`if (bm.is_dual_plane || (only_always && !bm.percentile_always) || !bm.percentile_hit)`).

I originally thought that it was simply looping over the full `block_modes` array, but I see now that it's incrementing `block_mode_count` as it has valid blocks, so it *should* only be iterating over valid blocks. After further investigation, the issue looks like `percentile_always` and `percentile_hit` aren't always being initialized. Valgrind no longer complains when initializing these to false with an added `else` clause in `construct_block_size_descriptor_2d()`, so I'll update the PR to use that fix.

solidpixel:
@akb825 Thanks - that one looks good to me.

Issue #219, opened 2021-02-21
Investigate per-block size tuning parameter defaults


solidpixel:
The current codec has global tuning parameters which are used for all block sizes. In practice this is inefficient - the choices you make for a 4x4 block cannot be the same as the choices you make for a 12x12 block because the block characteristics and available bit rate are just so different. I would propose splitting this into bands and having tuning parameters per band.

Initial thinking:

* High bitrate: 4x4, 5x4 and 5x5 (8bpp - 5.12bpp)
* Mid bitrate: 6x5, 6x6, and 8x5 (4.27bpp - 3.2bpp)
* Low bitrate: 8x6, 10x5, 10x6, and 8x8 (2.67 bpp - 2.0 bpp)
* Ultra-low bitrate: Everything else (1.6 bpp - 0.89 bpp)

... but in reality the major choices here are "high bitrate" vs "everything else" as only the high-bitrate options can really make full use of lots of partitions and dual weight planes with any fine detail accuracy.

solidpixel:
First implementation of this splits into three bands, based on some experimentation. See PR #249.

High bitrate: 4x4 to 5x4 (8bpp - 6.4bpp)
Mid bitrate: 5x5 to 8x6 (5.12bpp - 2.67bpp)
Low bitrate: 8x8 to 12x12 (2 bpp - 0.89 bpp)

PR #249, opened 2021-04-19
Add bitrate-specific presets

solidpixel:
Add presets that vary based on block size. Sensible presets for a 2bpp texture are very different to presets for a 8bpp texture, so this change allow us to vary presets in three bitrate bands. Implements #219.

* High bitrate band for `x < 25` texels (4x4 through to 5x4 blocks)
* Mid bitrate band for `25 <= x < 64` (5x5 though to 8x6 blocks)
* Low bitrate band `64 <= x` texels (8x8 blocks and up)

The most impactful setting has been dropping four partition searches from some searches presets, which significantly impacts performance for mid and low bitrate blocks where there is rarely enough bitrate to justify the attempt. To keep average PSNR up other settings have been nudged up to compensate. 

Nightly sweep of this gives (this is relative to 2.5, so not all of the performance uplift comes from this):

relative-main-vs-2 x


Source/astcenccli_toplevel_help.cpp
```diff
@@ -300,15 +300,18 @@ ADVANCED COMPRESSION
        ------------------------------------
 
        These options provide low-level control of the codec heuristics that
-       drive the performance-quality trade off.
+       drive the performance-quality trade off. The presets vary by block
+       bitrate; the recommended starting pioint for a 4x4 block is very
```
aras-p, line 304: `pioint` :)

aras-p:
The results look really nice!

solidpixel:
There is some overlap with this and the new heuristic culling, so for now I've merge the infrastructure for this but reset the settings so all bitrates are using the same. Tuning up the settings will be done separately.

PR #266, opened 2021-06-22
Support building macOS universal binaries

wasimabbas-arm:
If CMAKE_OSX_ARCHITECTURES is specified, it will be used to turn UNIVERSAL_BINARY to ON
otherwise its OFF by defaul


Source/cmake_core.cmake
```diff
@@ -136,68 +142,72 @@ macro(astcenc_set_properties NAME)
...
-        target_compile_options(${NAME}
-            PRIVATE
-                $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-mavx2 -mpopcnt -mf16c>
-                $<$<CXX_COMPILER_ID:MSVC>:/arch:AVX2>)
-
     elseif(${ISA_SIMD} MATCHES "native")
         target_compile_options(${NAME}
             PRIVATE
-                $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-march=native>)
+                $<$<CXX_COMPILER_ID:GNU>: -march=native>)
```
solidpixel, line 208: This breaks clang on Linux.

CMakeLists.txt
```diff
@@ -20,7 +20,20 @@ cmake_minimum_required(VERSION 3.15)
...
+project(astcencoder VERSION 3.0.0)
+
+option(UNIVERSAL_BUILD "Build multi-architecture universal binary [only supported on MacOS]" OFF)
+
+if (CMAKE_OSX_ARCHITECTURES)
+    message(STATUS "Building universal binary for ${CMAKE_OSX_ARCHITECTURES}")
+    set(UNIVERSAL_BUILD ON)
+endif()
+
+if (${UNIVERSAL_BUILD})
```
solidpixel, line 32: Do we need UNIVERSAL_BUILD as a variable? Can't we just infer it from the fact that CMAKE_OSX_ARCHITECTURES contains more than one architecture?

CMakeLists.txt
```diff
@@ -20,7 +20,20 @@ cmake_minimum_required(VERSION 3.15)
...
+    if (NOT CMAKE_OSX_ARCHITECTURES)
+        message(FATAL_ERROR "UNIVERSAL_BUILD requires -DCMAKE_OSX_ARCHITECTURES=\"arch1;arch2;...\"")
```
solidpixel, line 34: Would be a more useful message if we actually list real architectures. Currently the user needs to know the right settings for x86 and Arm, the message should just tell them.

CMakeLists.txt
```diff
@@ -20,7 +20,20 @@ cmake_minimum_required(VERSION 3.15)
...
+    endif()
+endif()
```
solidpixel, line 36: Builds currently still require the ARCH setting on the configure line, which doesn't make sense. (Line 51)

solidpixel (changes requested):

CMakeLists.txt
solidpixel: The current code no longer lets us build different SIMD binaries for macOS. It relies on CMAKE_OSX_ARCHITECTURES to select the targets, and the output is always a universal binary called `astcenc`. This is a regression for building distributable binaries; we need the ability to build all three x64 variants for binary releases (with different file names, as per the non-macOS behavior).

CMakeLists.txt
```diff
@@ -20,7 +20,20 @@ cmake_minimum_required(VERSION 3.15)
 cmake_policy(SET CMP0069 NEW)  # LTO support
 cmake_policy(SET CMP0091 NEW)  # MSVC runtime support
 
-project(astcenc VERSION 3.0.0)
+project(astcencoder VERSION 3.0.0)
+
+option(UNIVERSAL_BUILD "Build multi-architecture universal binary [only supported on MacOS]" OFF)
```
solidpixel, line 25: s/MacOS/macOS/

solidpixel:
Other work needed once the design settles down -  Update Docs/Building.md with the new instructions.

CMakeLists.txt
wasimabbas-arm: Fixed in https://github.com/ARM-software/astc-encoder/pull/266/commits/1ef2b84a31901c36f85479527cfe48d8365546a1

CMakeLists.txt
wasimabbas-arm: Fixed in https://github.com/ARM-software/astc-encoder/pull/266/commits/1ef2b84a31901c36f85479527cfe48d8365546a1

Source/UnitTest/cmake_core.cmake
```diff
@@ -42,77 +48,85 @@ target_compile_options(test-simd-${ISA_SIMD}
...
-# Set up configuration for SIMD ISA builds
-if(${ISA_SIMD} MATCHES "none")
-    target_compile_definitions(test-simd-${ISA_SIMD}
-        PRIVATE
-            ASTCENC_NEON=0
-            ASTCENC_SSE=0
-            ASTCENC_AVX=0
-            ASTCENC_POPCNT=0
-            ASTCENC_F16C=0)
+if (NOT ${UNIVERSAL_BUILD})
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/ARM-software/astc-encoder/pull/266/commits/1ef2b84a31901c36f85479527cfe48d8365546a1

Source/cmake_core.cmake
wasimabbas-arm: Fixed in https://github.com/ARM-software/astc-encoder/pull/266/commits/1ef2b84a31901c36f85479527cfe48d8365546a1

CMakeLists.txt
solidpixel: > `cmake -H. -Bbuild -DISA_AVX2=ON -DISA_SSE2=ON` should still work

Ack. Was begin dense last night =)

For `UNIVERSAL_BUILD` my main question was really whether we need it to be exposed as a user-visible `option()`. The user must set `CMAKE_OSX_ARCHITECTURES` for macOS which always implies `UNIVERSAL_BUILD`, so as a user-visible option having both seems redundant. Just turn `UNIVERSAL_BUILD` into an internal variable using `set()`?

CMakeLists.txt
solidpixel: > I was expecting this is assuming native Arm64 build. Because even in -DARCH=aarch64 case it wasn't cross compiling for aarch64 AFAIK

Cross-compiling works fine - the Jenkins macOS machine is x86-64. For macOS "it just works", for Linux you need to set `CXX` to the cross compiler you want to use.

solidpixel: > Agree it should be removed including the whole messages of backend - ON, backend - SKIPPED and backend - OFF what do you think?

The backend logging is useful, because CMake doesn't give useful errors if the user typos an option variable name, it just silently accepts it. Some type of feedback of what is actually getting built is required.

Source/cmake_core.cmake
```diff
@@ -136,68 +142,72 @@ macro(astcenc_set_properties NAME)
...
+                    ASTCENC_AVX=2
+                    ASTCENC_POPCNT=1
+                    ASTCENC_F16C=1)
+        endif()
+    endif()
 
-    elseif(${ISA_SIMD} MATCHES "sse2")
-        target_compile_definitions(${NAME}
+    if(${ISA_SIMD} MATCHES "avx2")
+        target_compile_options(${NAME}
```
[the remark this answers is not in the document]
solidpixel: Ack. I was worried that for SSE4.1 builds it wasn't guaranteed that -msse4.1 was getting set, but more obvious now that it is.

solidpixel (changes requested):

CMakeLists.txt
solidpixel: If we want to make it clear it's a hint, just stuff an "e.g." before "-DCMAKE...".

wasimabbas-arm: This whole section is gone after I remove `${UNIVERSAL_BUILD}` as a user visible option. User needs to know how to set `CMAKE_OSX_ARCHITECTURES`. Will add docs to clarify it instead.

CMakeLists.txt
wasimabbas-arm: User option removed in https://github.com/ARM-software/astc-encoder/pull/266/commits/b94c1f99323bc2d0a9bf63a6819ed27303671764

wasimabbas-arm:
> Other work needed once the design settles down - Update Docs/Building.md with the new instructions.

Updated docs in https://github.com/ARM-software/astc-encoder/pull/266/commits/dacae219bcad6826a81d4fd61fb51779fa6b359f

CMakeLists.txt
wasimabbas-arm: ARCH is now removed completely from cmake as well as jenkin scripts.

CMakeLists.txt
```diff
@@ -22,155 +22,143 @@ cmake_policy(SET CMP0091 NEW)  # MSVC runtime support
...
+endforeach()
 
-# Case when more than one x64 ISA and NEON is specified
-# This prevents something like ISA_AVX2 ISA_SSE41 ISA_NEON all specified
-if (${BUILD_NON_UNIVERSAL} AND ${ISA_NEON})
-        message(FATAL_ERROR "Conflicting ISAs specified. Only one of ISA_NONE|ISA_AVX2|ISA_SSE41|ISA_SSE2 is valid with ISA_NEON.")
-endif()
+# macOS builds
+if("${MACOS_BUILD}")
+    list(FIND CMAKE_OSX_ARCHITECTURES x86_64 IS_X64)
```
wasimabbas-arm, line 72: Unfortunately this doesn't work for xcode `-DCMAKE_OSX_ARCHITECTURES="\$(ARCHS_STANDARD)"` option.

CMakeLists.txt
```diff
@@ -22,155 +22,143 @@ cmake_policy(SET CMP0091 NEW)  # MSVC runtime support
...
 set(UNIVERSAL_BUILD OFF)
-set(X64_ISA_COUNT 0)
+set(MACOS_BUILD OFF)
+set(MACOS_ARCH_LEN 0)
 
-set(CONFIGS ${ISA_NONE} ${ISA_NATIVE} ${ISA_AVX2} ${ISA_SSE41} ${ISA_SSE2})
+# Preflight for some macOS-specific build options
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    set(MACOS_BUILD ON)
+    list(LENGTH CMAKE_OSX_ARCHITECTURES MACOS_ARCH_LEN)
```
wasimabbas-arm, line 49: You can't get this untill xcode starts the actual build, way later after the configure step unfortunately.

PR #274, opened 2021-07-01
Re-add native ISA support

solidpixel:
This PR re-adds support for a native ISA build, which becomes the default build if no specific options are set. The native ISA build will use the default compiler settings and compile-time detection of what level of SIMD is supported, rather than forcing it via defines. This works better with macOS builds, as it will dynamically cope with changing CPU architectures.


Source/CMakeLists.txt
```diff
@@ -37,8 +37,8 @@ if(${UNIVERSAL_BUILD})
     endif()
     include(cmake_core.cmake)
 else()
-    set(ARTEFACTS none neon avx2 sse4.1 sse2)
-    set(CONFIGS ${ISA_NONE} ${ISA_NEON} ${ISA_AVX2} ${ISA_SSE41} ${ISA_SSE2})
+    set(ARTEFACTS native none neon avx2 sse4.1 sse2)
```
wasimabbas-arm, line 40: There is one list like this in `Source/UnitTest/CMakeLists.txt` as well. Is that intentionally left out?

solidpixel: Good spot.

wasimabbas-arm (approved):

PR #276, opened 2021-07-12
Add a minimal example of using astcenc library API

solidpixel:


Utils/Example/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+cmake_minimum_required(VERSION 3.15)
+include(ExternalProject)
+
+project(astcencoder_example VERSION 1.0.0)
+
+# Command line configuration
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
```
wasimabbas-arm, line 28: This is not necessary.

Utils/Example/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+# Add the external project and pull out the project directories we need
+
+# Add e.g. "CMAKE_CACHE_ARGS -DISA_SSE41:String=ON" to ExternalProject_Add to
+# set configure settings usually found on the astcenc configure command line.
+# Remember that changing ISA changes the name of the library binary produced.
+
+ExternalProject_Add(astcencoder
```
wasimabbas-arm, line 41: I like `ExternalProject` command but I haven't seen much use of this in cmake projects. Usually an `add_subdirectory` is the most common way of using TPIP. With external project every time you want to do a clean rebuild, you nuke the whole download and start again. With subdirectory, the dependency needs to be a "sub-directory" for that to work but all of the following explicit methods like `target_include_directories/target_link_directories` goes away.

Utils/Example/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+        ${SOURCE_DIR}/Source)
+
+# ... with astcencoder Binary dir on the library path and as a library dep
+target_link_directories(astcenc_example
+    PRIVATE
+        ${BINARY_DIR}/Source)
+
+target_link_libraries(astcenc_example
+    PRIVATE
+        astcenc-native-static)
```
wasimabbas-arm, line 71: This is the tricky bit. Any advice on how this target name should be generated. I was thinking perhaps the build system can set a variable in parent scope to the name its trying to build and clients can pick it up.

Utils/Example/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations
+#  under the License.
+#  ----------------------------------------------------------------------------
+
+# CMake configuration
+cmake_minimum_required(VERSION 3.15)
+include(ExternalProject)
+
+project(astcencoder_example VERSION 1.0.0)
```
wasimabbas-arm, line 22: Some advice on which options to turn ON/OFF would be nice. Specially `-DCLI=OFF` in this case.

Utils/Example/astc_api_example.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+	// ------------------------------------------------------------------------
+	// Compress the image
+	astcenc_image image;
+	image.dim_x = image_x;
+	image.dim_y = image_y;
+	image.dim_z = 1;
+	image.data_type = ASTCENC_TYPE_U8;
+	uint8_t* slices = image_data;
+	image.data = reinterpret_cast<void**>(&slices);
```
wasimabbas-arm, line 115: Don't think slices is needed here. Just cast `image_data` directly?

Utils/Example/CMakeLists.txt
abbaswasim: This is resolved now.

Utils/Example/CMakeLists.txt
solidpixel: Might add that as a separate example - I can see both being useful to have.

Utils/Example/astc_api_example.cpp
solidpixel: It expects an array of pointers to slices, so the extra indirection is needed.

Issue #180, opened 2020-11-17
Consider adding a sampling swizzle (-ssw) option or auto-correcting channel weighting


solidpixel:
Today we have encoding swizzles (`-esw`) and decoding swizzles (`-dsw`), but nothing as an obvious partner to `-esw` which tells the compressor which channels are actually going be sampled from. 

For two component data, compressed using `-esw rrrg` to map the data to the two component LA endpoint, the compressor will still treats this as 4 component data in the error metrics. The red channel in the original data therefore get three times the significance in terms of weighting than the original green channel, which is rarely the desired behavior. 

You _can_ specify the correct behavior using the `-cw` option, e.g. `-cw 1 0 0 1` for this case which indicates that the application is only going to sample though `.ra`, so `.gb` can be discounted from error metrics. However it's not an obvious pairing with `-esw`, so the user needs to join the dots. Two proposals based on the current behavior:

* Add a new sampling swizzle option `-ssw` which enumerates the channels being sampled from. Other channels have their error weight forced to zero. This doesn't do anything that `-cw` can't do, but it's a nicer syntax, and a more obvious pairing with `-esw`.
* Ensure that the need to use either `-cw` or `-ssw` alongside `-esw` is clearly documented.

A third proposal is to implicitly determine default error weightings if we change the encoder to guarantee certain endpoints. For example, if the user passes in `-esw rrrg` data we could change the encoder to guarantee that only `L1` and `LA` endpoints would be used (we want to do this for performance anyway - see #172). If we have this guarantee then we know it doesn't matter which of the `.rgb` channels are sampled from, and we could automatically adjust so "rgb" and "a" have the same weight in the error calculations. This isn't safe without the guarantee; the compressor sometimes chooses RGB endpoints even for L data, just because the quant errors happen to round things the right way ...

solidpixel:
Temporary solution - I've added a note to the `-help` text in 8180b0f.

solidpixel:
Thinking this though I think that two channel data is the only case where this is a problem, as long a the user chooses sensible channel mappings.

- 1 channel data needs to be encoded as e.g. `rrr1`. Even if `r` is weighted three times more heavily there is nothing it's being traded off against - the constant 1 alpha value is free.
- 3 channel data needs to be encoded as e.g. `rgb1`. All three channels have equal weighting, the constant 1 alpha value is free.
- 4 channel data needs to be encoded as e.g. `rgba`. All four channels have equal weighting.

[...]

PR #307, opened 2022-02-19
Remove per-texel per-component error weighting

solidpixel:
## Summary 

Using per-texel per-component error weighting puts a lot of code and data lookups in the main compression path, costing up to 15% performance for the larger block sizes. This PR removes the ability to set a per-texel error weight for each color component, replacing it with a block-wide per-component weight instead. 

This is much faster and gives more scope for future optimizations, but removes some power-user configurability options from the compressor. The loss of functionality for use cases where power-users are manually using the settings to fix up compression artefacts (e.g. deblocking) can be offset by power-users spending some of the gained performance to increase search quality instead.

## Lost functionality:

* The `-cw` option to manually set a channel error-weight is no longer supported. 
* The `-b` option to reduce block artefacts is no longer supported. Users with a block artefact problem can use a higher search quality instead.
* The `-v` and `-va` options to set a per-texel error weight function based on statistical properties of a local region are no longer supported.
* The `-mask` and `-perceptual` options were based on the above, and are no longer supported.
* The `-a` option to set a per-texel error weight function based on the alpha in a local region is no longer supported. This also removes support for the RDO replacement for constant zero-alpha with no neighbouring non-zero-alpha, as we needed the alpha-radius functionality to determine if it was safe to replace a block under filtering. We could re-add the RDO support a different (cheaper) way in future ...
* The `-rgbm` mode was removed, but could now be readded (work in progress).

## Gained functionality:

* Internal block-wide channel weights are not exposed to the user, but are set automatically for any L+A blocks (including `-normal` textures) to correct error significance for lane replication. This also solves issue #180.

## Image quality impacts:

* The compressor was using per-texel weights to supress error contributions for texels that were outside of the image. This allowed only the in-image texels to be used to determine what a good encoding was.  Now all texels are used, minor image quality losses around the edges of images that are not an exact multiple of the block size are to be expected.
* The per-texel channel weights were primarily used for correcting the weighting for replicated RGB lanes in `-normal` and 2-component L+A material textures. The new per-block channel weight scheme is only a partial replacement, and does not re-add support everywhere that the original would have used it. This causes a minor drop in image quality, but this has been offset by spending some of the performance gained on a deeper search. Average normal quality has therefore improved.
* All HDR textures were using per-texel error weights based on a derivative of the LNS curve, which is no longer possible. Initial analysis shows that removing this has actually helped image quality, with the HDRIHaven test set showing 0.25 dB mPSNR improvements.

## Still to investigate:

* The `-cw` mode was removed, but could now be readded using block-wide error weighting.
* The `-rgbm` mode was removed, but could now be readded using block-wide error weighting.
* The `-a` mode for alpha-weighting removed, but could now be readded tactically in some places (not an exact match to the original). 
* Simple color perceptual (weight by luminance contribution), could now be readded using block-wide error weighting.
* Updating test reference scores to be based on these changes.
* Updating general markdown docs (if any).
* Restructuring the commits in this PR so it can be merged more sensibly as a set of incremental changes rather than a single big-bang commit (mostly to help bisect any image quality issues which might arise in future).


solidpixel:
Abandoned; replaced with the lighter-weight version in #308

PR #310, opened 2022-02-22
Convert Windows CI builds to use ClangCL

solidpixel:


solidpixel:
Converting to draft, as the Windows VM changes not yet in the live instance.

PR #311, opened 2022-03-10
Defragment partition tables to store only useful entries

solidpixel:
This PR defragments the partition tables, using a similar remapping scheme to block modes to turn a physical partition index used in the raw coding into the packed array index. 

Like block mode repacking, this approach only keeps entries that `astcenc` will use itself, and will drop otherwise legal partitions to avoid fragmenting memory. Therefore, enabling this repacking requires the context to be created with `ASTCENC_FLG_SELF_DECOMPRESS_ONLY`, as legal images created with other compressors may fail to decompress.

This change also introduces partition partition table initialization, based on the `tune_max_partitions` setting for the current compressor configuration. This helps reduce context creation time for -fast and -fastest modes, where we only need the tables for 1 and 2 partitions.


solidpixel:
Currently standalone decompression (`-dl`) is failing, but decompression during a round-trip (`-tl`) is working. Looks like we end up with a bad partitioning; decompression using the 3.4 binary works, so the compression pass looks OK. 

Reproducer:

`python3 ./Test/astc_test_functional.py CLIPTest.test_image_quality_stability`

PR #326, opened 2022-04-29
Tidy flags code sequence

halx99:


solidpixel:
Sorry, I'm not willing to accept this one. 

Flags don't need to be tidy - users should be using the symbolic values, and the actual number is only ever read by the machine. 

Changing this causes a binary ABI break so any user code will break if they fail to update the header, and will get fun bugs even if they recompile if their source code happens to be using the numeric values rather than the symbolic ones.

Binary ABI stability is more important than a cosmetic change here.

halx99:
ok, maybe move follow const to last is better
```cpp
static const unsigned int ASTCENC_FLG_MAP_RGBM             = 1 << 6;
```

Because when I try to add a new flags, I forget read all flags, and simply add a new flags same with `ASTCENC_FLG_MAP_RGBM             ` bug me some times.

solidpixel:
Yeah, reordering them in the header would be OK, just don't change the values. Note there can be holes in the values where old flags have been retired.

halx99:
Done by new one: #327

PR #336, opened 2022-05-19
Defer color packing to physical layer

solidpixel:
### Description

During compression the compressor currently cycles endpoint candidates through cycles of `pack` and `unpack`, even though the packed form is only actually needed for the physical layer encoding. 

This PR proposes deferring full color packing until the physical layer, so the encoding trials can simply worry about quantization rather than bitstream encoding details.

### Status

The branch is now deferring packing to the physical layer, which removes one table lookup from pack and unpack at the expense of adding an extra lookup table. This increases the ROM size due to the LUT, but decreases code size.  There is no significant performance benefit here yet, but this should unlock other optimizations.


solidpixel:
Rejecting - doesn't give any significant performance boost, and complicates the code quite a bit.

PR #342, opened 2022-06-08
Fixing build failure on windows arm64

Honeybunch:
As part of an ongoing saga in this PR: https://github.com/KhronosGroup/KTX-Software/pull/582

It turns out this project fails to build when targeting arm64 for windows. 

You can repro the failure yourself with:
* `cmake -B build-msvc-arm -G "Visual Studio 17 2022" -Aarm64 -DISA_NEON=ON`
* `cmake --build build-msvc-arm`

The issue comes down to how msvc handles NEON intrinsics. Since `uint32x4_t` is a union in msvc, I decided to override `operator[]` on vmask to reduce the footprint of the workaround.

### DIFF
--- a/Source/astcenc_vecmathlib_neon_4.h
+++ b/Source/astcenc_vecmathlib_neon_4.h
@@ -204,7 +204,7 @@ struct vint4
 	{
 		uint32x2_t t8 {};
 		// Cast is safe - NEON loads are allowed to be unaligned
-		t8 = vld1_lane_u32((const uint32_t*)p, t8, 0);
+		t8 = vld1_lane_u32(reinterpret_cast<const uint32_t*>(p), t8, 0);
 		uint16x4_t t16 = vget_low_u16(vmovl_u8(vreinterpret_u8_u32(t8)));
 		m = vreinterpretq_s32_u32(vmovl_u16(t16));
 	}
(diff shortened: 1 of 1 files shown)


Source/astcenc_vecmathlib_neon_4.h
```diff
@@ -346,6 +346,16 @@ struct vmask4
 		m = vreinterpretq_u32_s32(ms);
 	}
 
+	ASTCENC_SIMD_INLINE uint32_t operator[] (int32_t i)
+	{
+		// Workaround for how msvc handles intrinsic types
+#if defined(_MSC_VER)
+		return m.n128_u32[i];
+#else
+		return m[i];
```
solidpixel, line 355: Overloading the array accessors in ACLE has some weird implementation-defined behavior around endianess if you configure a big-endian Arm system. Rather than overloading "[]" I'd prefer this to provide the same "lane<id>()" function we have for the vint and vfloat types.

Honeybunch: Good to know! This seems cleaner overall anyway. 

Also threw in a small fix for this warning: `astc-encoder\Source\astcenc_vecmathlib_neon_4.h(885,37): warning C4244: 'return': conversion from 'int' to 'uint16_t', possible loss of data`

solidpixel (approved):

solidpixel:
@Honeybunch Thanks for putting this together - much appreciated.

PR #385, opened 2022-11-10
Defer color packing and scrambling

solidpixel:
This change moves packing and scrambling of quantized colors for the BISE bitstream to the physical layer. The symbolic layer of the codec always deals with colors in linear order in the 0-255 range, with sparse values if quantized. This matches a similar change we made for weights in an earlier release.

This improves performance by ~1%, at the expense of an extra set of data tables. One upside is that it may be possible to avoid the data tables if we can programmatically correct for the slightly weird rounding caused by the BISE decode.


solidpixel:
This change moves packing and scrambling of quantized colors for the BISE bitstream to the physical layer. The symbolic layer of the codec always deals with colors in linear order in the 0-255 range, with sparse values if quantized. This matches a similar change we made for weights in an earlier release.

This improves performance by ~1%, at the expense of an extra set of data tables, although the ROM cost is offset by making one of the other sets of tables in to a table of variable length tables, instead of a regular 2D array.

PR #388, opened 2022-12-02
Disable blue contraction at quant level 256

castano:
At quantization level 256 there's no actual quantization happening, so blue contraction should not provide any benefit in that case. In fact, allowing blue contraction at that level seems to increase the error slightly. These are the results of running the encoder on the kodak image set:

```
new:
  fastest   RMSE = 2.9423
  fast      RMSE = 2.7559
  medium    RMSE = 2.6331

original:
  fastest   RMSE = 2.9478
  fast      RMSE = 2.7692
  medium    RMSE = 2.6470
```

I haven't evaluated performance. I would expect a slight increase, but may not be significant enough to be noticeable.

### DIFF
--- a/Source/astcenc_color_quantize.cpp
+++ b/Source/astcenc_color_quantize.cpp
@@ -1973,7 +1973,7 @@ uint8_t pack_color_endpoints(
 				break;
 			}
 		}
-		if (try_quantize_rgb_blue_contract(color0, color1, output, quant_level))
+		if (quant_level < QUANT_256 && try_quantize_rgb_blue_contract(color0, color1, output, quant_level))
 		{
 			retval = FMT_RGB;
 			break;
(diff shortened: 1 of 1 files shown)


solidpixel:
Thank you =)

Summary of running through testing here:
* IQ is nicely improved for 4x4; not much change outside of that. One of the Khronos test images improves by over 0.2 dB, which is awesome. 
* Performance change is in the noise margin.

castano:
Glad to hear that it helps! Makes me wonder if blue contraction is sometimes increasing the error even though it helps in the average case, and whether that could be predicted somehow (without checking both and picking the most accurate representation).

BTW, are the Khronos test images publicly available? Would love to add them to my test suite.

solidpixel:
> BTW, are the Khronos test images publicly available? 

I use the ones in the KTXTools repo, which are available on GitHub here:
* https://github.com/KhronosGroup/KTX-Software/tree/master/tests/srcimages

> whether that could be predicted somehow

I'll mockup a a brute-force try both and see if that helps to see what's on the table.

solidpixel:
> I'll mockup a a brute-force try both and see if that helps to see what's on the table.

Didn't seem to make much difference.

PR #393, opened 2023-01-19
mathlib: Remove incomplete support for SSE3 which assumed SSSE3

akien-mga:
`_mm_shuffle_epi8` requires SSSE3 so the check on `ASTCENC_SSE >= 30` is
too lax and would fail if `__SSE3__` is supported, but not `__SSSE3__`.

The only supported configurations are SSE2, SSE4.1, and AVX2, so as
discussed in #393 we drop the SSE3 checks and require SSE4.1 instead.

Downstream issue: https://github.com/godotengine/godot/issues/71700

Confirmed the bug by building astcenc with MinGW-GCC on Linux with `-msse3`, and confirming that it defines `__SSE3__` but not `__SSSE3__`.
Adding `-mssse3`, or my patch, fixes the issue.

### Notes

- ~I'm not very knowledgeable on SIMD stuff. I don't know if "31" is a good index for `SSSE3` here.~
- ~I updated all `>= 30` checks to `>= 31` assuming that they need to be kept in sync. Please advise if I should change some back.~
- Updated to require SSE4.1 for this code instead, as discussed below.

### DIFF
--- a/Source/astcenc_mathlib.h
+++ b/Source/astcenc_mathlib.h
@@ -48,8 +48,6 @@
     #define ASTCENC_SSE 42
   #elif defined(__SSE4_1__)
     #define ASTCENC_SSE 41
-  #elif defined(__SSE3__)
-    #define ASTCENC_SSE 30
   #elif defined(__SSE2__)
     #define ASTCENC_SSE 20
   #else
--- a/Source/astcenc_vecmathlib_sse_4.h
+++ b/Source/astcenc_vecmathlib_sse_4.h
@@ -1046,7 +1046,7 @@ ASTCENC_SIMD_INLINE void vtable_prepare(vint4 t0, vint4& t0p)
  */
 ASTCENC_SIMD_INLINE void vtable_prepare(vint4 t0, vint4 t1, vint4& t0p, vint4& t1p)
 {
-#if ASTCENC_SSE >= 30
+#if ASTCENC_SSE >= 41
 	t0p = t0;
 	t1p = t0 ^ t1;
 #else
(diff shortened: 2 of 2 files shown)


solidpixel:
Thanks for the PR - will double check the SIMD support for this one.

akien-mga:
I don't think we need SSE3/SSSE3 specifically, it just happens that apparently WSL2 (at least for @kuhnchris) defaults builds to enable SSE3 but not SSSE3, so the current checks broke.

Moving your `>= 30` checks for APIs that need SSSE3 to `>= 41` would be perfectly fine with me.

solidpixel:
I would prefer not to support it if you don't need it (nearly all hardware sold in the last 5 or 6 years has SSE4.1), so I think my preferred fix for this is just to delete this entry in the auto-detection ladder:

```
  #elif defined(__SSE3__)
    #define ASTCENC_SSE 30
```

... so it falls back to the SSE2 codepath if 4.1 isn't available.

kuhnchris:
Just to point out: this is the default mingw install from `scoop` running directly on windows, the report I sent to @akien-mga was sent from WSL, but used the gcc from the mingw install - so this shouldn't make much of a difference. Maybe it's just that the default flags by mingw are weird? Because apparently the CPU *would* support SSSE3, the compiler just chose not to include it. But I'm far not knowledgable enough to know how/why gcc is providing flags so I'll leave it up to you guys, just let me know if I should test something regarding this issue. :-)

The CPU in question:
image

akien-mga:
Updated as suggested. I still changed `>= 30` to `>= 41` since there's no point in the `30` threshold anymore (and it was incorrect since it needs SSSE3).

solidpixel (approved):

solidpixel:
Thank you, much appreciated.

PR #398, opened 2023-02-07
Output error message to stderr

halx99:


solidpixel (approved):
Looks good - will try some local builds shortly.

Source/astcenccli_toplevel.cpp
```diff
@@ -30,6 +30,8 @@
 #include <vector>
 #include <memory>
 
+#define print_error(format, ...) fprintf(stderr, format, ##__VA_ARGS__)
```
solidpixel, line 33: String pasting `,` with `__VA_ARGS__` is a GNU extension - fails on Clang (not tested MSVC). `__VA_OPT__` isn't available until C++20, which is too new for the time being.

solidpixel: Will look for a cleaner way to fix this than making every error without args print via a "%s" template.

solidpixel (approved):

PR #400, opened 2023-02-21
Implement veneer for ISA compat checks

### LINKED ISSUE #384: `astcenc_config_init()` crashes when astc-encoder is compiled with AVX2 instructions on CPUs that don't support AVX2
In theory, astc-encoder will [check for AVX2 support](https://github.com/ARM-software/astc-encoder/blob/026235b65e3480704599ec865d873dc0797b1156/Source/astcenc_entry.cpp#L685) during `astcenc_context_alloc()` and fail gracefully if the ISA isn't supported.

However, when AVX2 is enabled, it seems like the compiler uses AVX2 instructions in `astcenc_config_init()`, specifically [this line](https://github.com/ARM-software/astc-encoder/blob/026235b65e3480704599ec865d873dc0797b1156/Source/astcenc_entry.cpp#L533).

As a result, it crashes with `EXCEPTION_ILLEGAL_INSTRUCTION` even before the library has a chance to gracefully fail.

solidpixel:
This PR implements a small veneer on the core library that is designed to be compiled using a base ISA which is guaranteed to be supported on the target architecture. This ensures that we can safely perform the ISA compatibility check in the library without failing with a SIGILL.

Fixes #384.

### DIFF
--- a/Source/astcenc_entry.cpp
+++ b/Source/astcenc_entry.cpp
@@ -158,48 +158,6 @@ static astcenc_error validate_cpu_float()
 }
 
 /**
- * @brief Validate CPU ISA support meets the requirements of this build of the library.
- *
- * Each library build is statically compiled for a particular set of CPU ISA features, such as the
- * SIMD support or other ISA extensions such as POPCNT. This function checks that the host CPU
- * actually supports everything this build needs.
- *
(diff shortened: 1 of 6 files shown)


solidpixel:
Rejecting - decided to move the compat check out of the core library.

PR #405, opened 2023-04-24
Enable building x86_64h slices on macOS

pthariensflame:
This additional x86_64 variant target only exists on macOS and represents a Haswell-or-later CPU—i.e., with AVX2 available.  It can be used as an additional slice of a universal binary alongside x86_64 and arm64, enabling a single universal binary to get maximal performance on all supported Macs, since the system will automatically use the x86_64h slice if it's available and the CPU supports it.


solidpixel:
Thanks - will take a look.

pthariensflame:
Notes on the current state of this PR:  the build works, and indeed produces a universal binary that uses only up to SSE4.1 instructions in the `x86_64` slice while using AVX2 instructions in the `x86_64h` slice (verified by using `lipo` to extract slices and then [`elfx86exts`](https://github.com/pkgw/elfx86exts) on each slice); however, I'm not convinced that the build actually configured the `x86_64h` slice to use AVX2 in the source code, rather than just automatically using some instructions through autovectorization alone.  That's why this is a draft.

solidpixel:
Sorry for the slow progress - had to go delving in the office to find a older machine I can actually test this on - mine is an M1, so can't test any AVX code. I found one, so hope to pick this up next week.

solidpixel:
I think this might be working, but there are some cosmetic issues, for example:

```
lipo ./bin/astcenc -extract x86_64 -output astcenc-x64
./astcenc-x64 -v
lipo ./bin/astcenc -extract x86_64h -output astcenc-x64h
./astcenc-x64h -v
```

... sees to show the x64h build picking up the specified SSE build variant, not AVX2. 

In general relying on _any_ preprocessor state for the universal builds is problematic because the same preprocessor state is used by all of the build targets, irrespective of architecture. I do wonder if the best fix for this is to make single architecture binaries, and then use `lipo` to merge them into a single universal binary at the end.

solidpixel:
Closing this one, but please see the follow-on PR #423 for an implementation which should work.

PR #409, opened 2023-05-15
Add namespacing to CMake interface

solidpixel:
This change adds namespacing to CMake variables to make the library easier to integrate.

* Less likely to have collisions that break other included CMake files if calling directly. 
* Related arguments show up together in the CMake config GUI.


Source/UnitTest/cmake_core.cmake
```diff
@@ -122,18 +122,18 @@ elseif(${ISA_SIMD} MATCHES "avx2")
...
 endif()
 
-target_link_libraries(${ASTC_TEST}
+target_link_libraries(${ASTCENC_TEST}
     PRIVATE
         gtest_main)
 
-add_test(NAME ${ASTC_TEST}
-         COMMAND ${ASTC_TEST})
+add_test(ASTCENC_TARGET_NAME ${ASTCENC_TEST}
```
wasimabbas-arm, line 136: This is a mistake. "NAME" is keyword in "add_test" it shouldn't be renamed.

wasimabbas-arm: Also how does the actual name shows up? looks like the "COMMAND" and "NAME" are the same.

wasimabbas-arm (changes requested):

Source/CMakeLists.txt
```diff
@@ -17,44 +17,44 @@
...
+    set(ASTCENC_CONFIGS ${ASTCENC_ISA_NATIVE} ${ASTCENC_ISA_NONE} ${ASTCENC_ISA_NEON} ${ASTCENC_ISA_AVX2} ${ASTCENC_ISA_SSE41} ${ASTCENC_ISA_SSE2})
+    list(LENGTH ASTCENC_ARTIFACTS ASTCENC_ARTIFACTS_LEN)
+    math(EXPR ASTCENC_ARTIFACTS_LEN "${ASTCENC_ARTIFACTS_LEN} - 1")
 
-    foreach(INDEX RANGE ${ARTIFACTS_LEN})
-        list(GET ARTIFACTS ${INDEX} ARTIFACT)
-        list(GET CONFIGS ${INDEX} CONFIG)
-        if(${CONFIG})
-            set(ISA_SIMD ${ARTIFACT})
+    foreach(ASTCENC_INDEX RANGE ${ASTCENC_ARTIFACTS_LEN})
```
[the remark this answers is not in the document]
solidpixel: Let's not introduce more typing if we can avoid - will remove.

wasimabbas-arm:
Approved.

PR #412, opened 2023-05-17
Use precise fp with explicit contraction control where possible

solidpixel:
VS2019 and older still require strict floating point due to the
older /fp:precise behavior implying contraction, which was improved
in VS2022.


Docs/ChangeLog-4x.md
```diff
@@ -14,13 +14,17 @@ clocked at 4.2 GHz, running `astcenc` using AVX2 and 6 threads.
...
 * **General:**
-  * **Bug-fix:** Invariant Clang and GCC builds now force `-ffp-model=strict`,
-    which is needed due to recent changes in compiler defaults. These builds
-    will be slower than the previous release, which was not guaranteed to be
-    invariant. Use a non-invariant build to recover lost performance if needed.
-  * **Bug-fix:** Non-invariant Clang and GCC builds now force
-    `-ffp-model=precise`, which is needed due to recent changes in compiler
-    defaults.
+  * **Bug-fix:** Invariant Clang and GCC builds now use `-ffp-model=precise`
+    with `-ffp-contract=off` which is needed due to restore invariance due to
```
wasimabbas-arm, line 18: minor "needed to" remove 'due'

wasimabbas-arm:
Would it make sense to abstract these options out to users? That way those who don't care about bit-precise results can avoid taking the perf hit.

solidpixel:
> Would it make sense to abstract these options out to users?

It's all hidden under the `-DNO_INVARIANCE=ON/OFF` build option in practice, so from that point of view there is no change other than the implementation catching up with the latest compiler changes.

solidpixel:
CI failure on GCC 7.5 not supporting `-ffp-model` option. Need to see what we can do there (either drop support for old compilers or handle version-specific option).

PR #417, opened 2023-06-02
Use more robust CMake compiler switch selection

solidpixel:
Using "Clang" in CMake generators for selecting command line switches is now an ambiguous compiler identifier, as both Clang and ClangCL report as "Clang" for `CMAKE_CXX_COMPILER_ID`. 

We need to use some combination of `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` and `CMAKE_CXX_COMPILER_ID` to uniquely identify the compiler so we can use the correct flavor of command line switches.

This PR also fixes buggy floating-point contraction handling in the VS2022 ClangCL compiler, which was breaking invariance.

### DIFF
--- a/Docs/ChangeLog-4x.md
+++ b/Docs/ChangeLog-4x.md
@@ -14,6 +14,8 @@ clocked at 4.2 GHz, running `astcenc` using AVX2 and 6 threads.
 The 4.5.0 release is a maintenance release with minor fixes and improvements.
 
 * **General:**
+  * **Bug-fix:** Improved handling compiler arguments in CMake, including
+    consistent use of MSVC-style command line arguments for ClangCL.
   * **Bug-fix:** Invariant Clang builds now use `-ffp-model=precise` with
     `-ffp-contract=off` which is needed to restore invariance due to recent
     changes in compiler defaults.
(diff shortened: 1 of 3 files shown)


solidpixel:
The CI builds failed with some compiler argument issues on Windows for VS2019 with SSE 4.1, and macOS with AVX2, so need to debug those.

MarkCallow (approved):
The changes look good to me.

Unfortunately in the case of MSVC vs ClangCL on Windows11/arm64 there is no reduction in the list of failing tests. I have yet to test on macOS and Linux.

solidpixel:
> Unfortunately in the case of MSVC vs ClangCL on Windows11/arm64 there is no reduction in the list of failing tests. I have yet to test on macOS and Linux

@MarkCallow Can you confirm the -Xclang -ffp-contract=off is getting picked up correctly for the Arm builds? I'll see if I can get hold of a machine to repro locally.

MarkCallow:
> @MarkCallow Can you confirm the `-Xclang -ffp-contract=off` is getting picked up correctly for the Arm builds? I'll see if I can get hold of a machine to repro locally.

It appears under Additional Options in the Visual Studio 2022 GUI. Worryingly it does not appear in the "Command Line" the GUI displays but I think that is an artifact of how VS displays options. I am having trouble finding any log of the build that shows the actual commands run so at present can't verify any further.

MarkCallow:
> I am having trouble finding any log

I found the log file. `-Xclang -ffp-contract=off` is definitely set in the compile commands.

PR #435, opened 2023-10-03
Reduce use of reinterpret_cast

solidpixel:
This PR reduces use of `reinterpret_cast` in the core codec to minimize strict-aliasing violations. 

It is assumed that `reinterpret_cast` in the SIMD intrinsics headers are fine, as the types are tagged in the compiler implementation as "may alias".


ktkachov-arm:
FWIW this looks good to me.

PR #440, opened 2023-11-22
Support using more than 64 cores on Windows

solidpixel:


solidpixel:
Confirmed uplift on 96 core/192 thread AMD EPYC, and get 2x uplift vs 64 threads.

PR #442, opened 2023-11-28
Make unit tests more robust across different architectures

roehling:
This PR deals with a few architectural idiosyncrasies that came up in Debian:

- The index lookups are sensitive to endianness
- Floating point equality comparisons are generally considered flaky, which is why GTest provides `EXPECT_FLOAT_EQ`. The i387 coprocessor with its internal 80 bit extended precision makes things especially awkward.

### DIFF
--- a/Source/UnitTest/test_simd.cpp
+++ b/Source/UnitTest/test_simd.cpp
@@ -1969,10 +1969,17 @@ TEST(vint4, vtable_8bt_32bi_32entry)
 
 	vint4 result = vtable_8bt_32bi(table0p, table1p, index);
 
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
 	EXPECT_EQ(result.lane<0>(),  3);
 	EXPECT_EQ(result.lane<1>(),  4);
 	EXPECT_EQ(result.lane<2>(),  7);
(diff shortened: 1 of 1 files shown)


solidpixel:
> The index lookups are sensitive to endianness

Ack, seems sensible to fix. Just want to make sure this is a portable way to fix the problem (we get some esoteric compilers and old compiler versions used in embedded systems).

> Floating point equality comparisons are generally considered flaky, which is why GTest provides EXPECT_FLOAT_EQ.

The main point of these tests is to ensure that the SIMD is actually bit-exact with the non-SIMD code on a specific CPU, rather than prove out any kind of floating point cross architecture, so in this case the exact comparison was somewhat deliberate. 

Is there actually a real use case for supporting i387 beyond it being an interesting piece of architectural history? We've long since dropped 32-bit platforms from our testing.

solidpixel:
I can do that, no probs.

solidpixel:
It too a while, but better late than never, here is a PR that adds proper big-endian support: #561.

I'm not relying on any compiler preprocessor values to auto-detect it, so it relies on the user requesting it as a CMake configure option.

solidpixel:
Closing now that BE support has been added via the other PR.

PR #445, opened 2023-12-03
Add error handling for pthread_create failing

solidpixel:


elfring:
How do you think about to use an output channel like “[stderr](https://en.wikipedia.org/wiki/Standard_streams#Standard_error_%28stderr%29 "Documentation")” for [special messages](https://github.com/ARM-software/astc-encoder/commit/bbb6c4704c890da292db1a55b27c89f72381866b#diff-c419ab6622349a838f96ce237f39f217b809e64959d61633f7eabb414f168b37R249-R256 "Update candidate: launch_threads()")? :thinking:

solidpixel:
We do for errors. Warnings are not errors.

elfring:
:thought_balloon: How many “warnings” would be sent to the channel “standard output” so far?

PR #455, opened 2024-02-27
Move STB & TinyEXR headers out of main source directory

jherico:
If a project incorporates astc-encoder as a submodule (as Vulkan-Samples [does](https://github.com/KhronosGroup/Vulkan-Samples/blob/8ee282fddc105272ae99a0cb35491ce8b0c8e3ae/.gitmodules#L37) ) then it's impossible to target the include folder for `astcenc.h` without also getting the STB and TinyEXR headers. 

This is problematic because if the downstream project is also using either of those modules directly (again, as Vulkan-Sample does with [STB](https://github.com/KhronosGroup/Vulkan-Samples/blob/8ee282fddc105272ae99a0cb35491ce8b0c8e3ae/.gitmodules#L7)) then it can create a potential version conflict, as which header a given downstream file gets depends entirely on the order of the project dependencies, which ends up creating hard to discover bugs.

This PR moves the non-ASTC source code into a `ThirdParty` folder in order to resolve this.  An alternative change would be to move the `astcenc.h` file out of `/Source` and into `/Include` so that it's clearly indicated as the public interface of the project.  However, that kind of change would impact other downstream users of the project who are also using this as an embedded project.


Source/astcenccli_image_external.cpp
```diff
@@ -66,9 +66,9 @@ static void astcenc_runtime_assert(bool condition)
     }
 }
 
-#include "stb_image.h"
-#include "stb_image_write.h"
-#include "tinyexr.h"
+#include "third_party/stb_image.h"
```
solidpixel, line 69: Please can you change the name of the directory from `third_party` to `ThirdParty` so this follows the existing convention in the project.

jherico: fixed

solidpixel:
I'll merge off a branch once it's gone through our CI.

jherico:
Whoops... sorry, I thought I'd updated the branch, but I'd forgotten to stage the changes.

solidpixel:
No worries - fixed on the branch =)

solidpixel:
Merged to main.

Issue #361, opened 2022-08-10
Investigate ASTC RDO


solidpixel:
There is some interest in rate distortion optimization to improve compressibility of ASTC textures in application package bundles for download. 

* https://github.com/KhronosGroup/KTX-Software/issues/588

### Approaches

The first option would be an in-situ RDO, where the output is still valid ASTC data. This is transparent to the application, which is a major advantage, but harder to pull off (especially given the packed variable-length scrambled BISE encoding used by ASTC). 

The second option would be a repack RDO, where we transcode into a light-weight packaging format. This is not transparent - the application must unpack at load time - but gives opportunity to systematically target inefficiency in the encoding. One particular strength of this approach is that we can use data tables that are built into the encoder/decoder rather than wasting bits based on dynamic tables built into the compressed stream. 

In both cases it can be an advantage to add an RDO mode to the compressor, where it knows that its output will be post-processed by an RDO compressor and so tries to constrain the raw encodings to values that are more amenable to RDO compression. 

### Inefficiencies we can target

* The block mode encoding is statically inefficient for small block sizes (e.g. most decimation modes are invalid for 4x4).
* The partition hash is statically inefficient, particularly for small block sizes (e.g. duplicate partitions, seeds that don't generate all N partitions, partitions that are not used because they are low value). This is probably an area for compressor restriction in an RDO mode to avoid using low value partitions. 
* The CEM modes are dynamically inefficient (e.g. never use HDR in an LDR texture, never use RGBA in a RGB texture).

## Other random ideas

* Does using blue-contract reduce compressibility during stream compression or reusability during repacking?
* Does using BISE scrambling reduce compressibility during stream compression or reusability during repacking?

YunHsiao:
Here's one possible way to do it:
https://github.com/ARM-software/astc-encoder/compare/4.7.0...YunHsiao:astc-encoder:rdo

PR #460, opened 2024-03-28
RDO Postprocess

YunHsiao:
This is one possible implementation for #361, which is based on Richard Geldreich's original [ert implementation](https://github.com/richgel999/bc7enc_rdo/blob/master/ert.cpp) (also subsequently used in KTX2 implementation for UASTC), with some interface tweaks that makes the solver more general and performant to integrate with.


YunHsiao:
I've reached a point where I think most of the work is solid, but I'd love some second opinions:
* Implemented as a separate step with an independent `ParallelManager`
* Disabled by default, with CLI interfaces closely resembles the relevant parameters in [KTX2](https://github.khronos.org/KTX-Software/libktx/structktxBasisParams.html)
* The diff function for trial blocks are routed to existing specialized ones for different planes & partitions, with a new one added for the constant symbolic block case
* HDR, swizzle, different weights should be working but I haven't looked much into it yet
* Not sure about how to handle the image borders where the block isn't necessarily full of valid texels

solidpixel:
Have you looked at image quality to bitrate trade offs, and how this compares to just using a larger ASTC block size to reduce bitrate?

MarkCallow:
> Have you looked at image quality to bitrate trade offs, and how this compares to just using a larger ASTC block size to reduce bitrate?

I'm not sure I understand the question. RDO is all about findign a good balance between quality (distortion) and (bit)rate. The intent is to reduce entropy between blocks to reduce the size after Deflate-style lossless compression. I wonder if there is a point at which a larger block size produces a smaller result than RDO+Deflate. It would be good to know. But keep in mind that the purpose of RDO+Deflate is to reduce file and transmission size. It does not affect the GPU memory used.

solidpixel:
> the purpose of RDO+Deflate is to reduce file and transmission size. 

Yes, but as you noted ASTC can do the same thing natively (just use a larger block size) which gives file, transmission, AND memory savings. There isn't much point adding this if it ends up the wrong side of the pareto frontier vs just using a smaller ASTC block size.

MarkCallow:
> pareto frontier

What is this?

> I'd still like the PR to include some analysis of the new feature and quality/bitrate which can be added to the documentation so we can explain when it should be used, and how to use it effectively.

Absolutely. From that we can determine if it is useful. We need to compare the zstd deflated file with and without RDO to determine its effectiveness.

solidpixel:
> > pareto frontier
> What is this?

Scatter plot whatever pair of metrics you care about for a compressor (quality-performance and quality-size are the two that tend to get used). The pareto frontier is the line made through the best compressor/configuration for a given quality or performance setting. 

For a new optimisation to be useful it needs to be above the frontier - e.g. giving better performance for the same quality, or visa versa. If you're behind the frontier, then there isn't any point - better options already exist. 
For a new configutation to be

YunHsiao:
Okay finally had some time to get to this... Although I wasn't able to make a full performance report, some simple test can still provide useful insights:

For a 2048x2048 test image, the following script is run:
```bash
./astcenc-avx2.exe -tl test.jpg 6x6.png 6x6 -medium -dimage
./astcenc-avx2.exe -tl test.jpg 6x6r.png 6x6 -medium -dimage -rdo-quality 1
7z a 6x6.7z 6x6.astc
7z a 6x6r.7z 6x6r.astc

./astcenc-avx2.exe -tl test.jpg 8x6.png 8x6 -medium -dimage
./astcenc-avx2.exe -tl test.jpg 8x6r.png 8x6 -medium -dimage -rdo-quality 1
7z a 8x6.7z 8x6.astc
7z a 8x6r.7z 8x6r.astc

./astcenc-avx2.exe -tl test.jpg 8x8.png 8x8 -medium -dimage
./astcenc-avx2.exe -tl test.jpg 8x8r.png 8x8 -medium -dimage -rdo-quality 1
7z a 8x8.7z 8x8.astc
7z a 8x8r.7z 8x8r.astc

./astcenc-avx2.exe -tl test.jpg 10x10.png 10x10 -medium -dimage
./astcenc-avx2.exe -tl test.jpg 10x10r.png 10x10 -medium -dimage -rdo-quality 1
7z a 10x10.7z 10x10.astc
7z a 10x10r.7z 10x10r.astc

./astcenc-avx2.exe -tl test.jpg 12x12.png 12x12 -medium -dimage
./astcenc-avx2.exe -tl test.jpg 12x12r.png 12x12 -medium -dimage -rdo-quality 1
7z a 12x12.7z 12x12.astc
7z a 12x12r.7z 12x12r.astc
```

The key results are as follows:

| Block Size | Non-RDO PSNR | RDO PSNR | Non-RDO Zip Size | RDO Zip Size |
|:----------:|:-------------:|:--------:|:-----------------:|:------------:| 
|    6x6     |    36.1594    | 34.7948  |       3491        |     2711     |
|    8x6     |    34.2204    | 33.3474  |       2552        |     2107     |
|    8x8     |    32.5107    | 31.9004  |       1919        |     1581     |
|   10x10    |    29.8623    | 29.5536  |       1253        |     1032     |
|   12x12    |    28.1800    | 27.9452  |        864        |     695      |

So I guess it may not be some ground-breaking optimization:
* 8x6 seems like a better option than 6x6-rdo considering RDO is usually quite time-consuming

But in many cases it still may provide interesting trade offs:
* 8x6-rdo results sits comfortably right between 8x6 & 8x8 (adjusting compression quality per-se wouldn't make this much of a difference)
* We can still push some boundaries on top of 12x12
* In our own use cases, comparing outputs with near-identical PSNRs (e.g. 8x6-thorough vs. 6x6-rdo), we tend to feel the RDO ones almost always yield better results, but of course this a subjective matter.

solidpixel:
Super, thanks. I'll try and take a look next week.

solidpixel:
> btw, apart from how well it affects package size, the encoding speed is currentlly a big problem ...

Agree. The original BC implementation has the same problem because it's a somewhat brute force approach to finding RDO matches. Suspect it's too slow to be usable at scale, given the marginal benefits over just choosing other ASTC block sizes.

YunHsiao:
Alright, like I said in the beginning this is only one possible approach to do this, it is not ideal but to some extent it perfectly works. I came for possible advices on the implementation and as far as it goes, it stood up to our production usages. @solidpixel Feel free to close this if this approach is not expected to be merged.

MarkCallow:
Please explain why/how it "perfectly works" and, if possible, give some more comparisons between this and choosing other block sizes.

YunHsiao:
It perfectly works bcs any invalid trial blocks are rejected aggressively.

In practice, on top of a reasonable block-size selection scheme, if the transmission size out-benefits the cook time required, then it is a perfectly valid optimization. For our unreal project, it can cut down ~500M package size with carefully selected compression parameters. Also with DDC the build is always incremental, so overall not that expensive.

solidpixel:
> @solidpixel Feel free to close this if this approach is not expected to be merged.

At the moment, I don't have any work schedule for astcenc, so I'm really only merging essentials for platform support, compiler support, and bug fixes. I'd like to add RDO support, but I'm just wary of adding something new to test and maintain at the moment.

richgel999:
> This is one possible implementation for #361, which is based on Richard Geldreich's original [ert implementation](https://github.com/richgel999/bc7enc_rdo/blob/master/ert.cpp) (also subsequently used in KTX2 implementation for UASTC), with some interface tweaks that makes the solver more general and performant to integrate with.

Glad to see this code is useful. I did test the idea with ASTC LDR (that's why the key ERT function supports user supplied block sizes). I've recently implemented an RDO ASTC HDR 6x6 solution, which just shipped in the Basis Universal lib. Here are some thoughts on RDO ASTC. I don't expect this to move the needle either way on this PR, this note is mostly for anyone who stumbles on it:

- Block artifacts at the larger block sizes can be very noticeable to the customer in ways that PSNR doesn't capture. PSNR isn't a reliable measure of perceived quality, especially at the larger block sizes. Even bad block artifacts at 4x4 can be very hard to see in practice, but block artifacts at 6x6 or larger can be so visible to make some encodes virtually useless. This is why we've switched to a weighted mixture of generalized SSIM+MSE. The generalized formula allows us to tune the SSIM metric to be more sensitive to mean error, to reduce block artifacts. So the above PSNR statistics aren't the whole story. 
- RDO ASTC gives the developer an extra valuable knob they can turn, without having to go to a larger ASTC block size and risk more noticeable block artifacts. At HDR this is even more of a risk because quality can degrade rapidly as the block size increases in size.
- The other major non-ASTC formats (BC1-7) already have RDO encoders available, but the idea has been underexplored with ASTC until now.
- The above transform (ERT) is a brute force match injection post-process that can be applied to the raw ASTC texture data bits. It doesn't impact the encoding at all. It can be optimized and improved in various ways. Dropping it into existing encoders should be straightforward.
- Because the raw ASTC texture data is being modified, it's very important to ensure the resulting distorted ASTC data is still 100% valid ASTC. I've found that each software decoder has varying degrees of tolerance for data which slightly deviates from the spec in some way. The SW decoders don't go out of their way to 100% validate the block data. I would be careful to ensure that the resulting modified ASTC texture data is decoded correctly on all the GPU's and software decoders. Just ensuring it decodes with one software decoder is likely not enough validation.
- The ERT in the code above can be improved to add SSIM, a block artifact-aware error metric, and endpoint/weight refinement, and better colorspaces such as Oklab or ICtCp etc.

PR #473, opened 2024-06-07
Vector library cleanup

solidpixel:
The astcenc vector library effectively implements two different class APIs:

* An explicit 4-wide API which is used via explicit 4-wide types (e.g. `vfloat4`) in the codec.
* A vector-length agnostic API, which is used as N-wide types in the codec (e.g. `vfloat`) in the codec, and where the width is resolved at compile time.

For historical reasons the classes that are only used as a VLA classes (e.g. `vfloat8` for AVX2) implement a lot of functionality which was inherited from the original 4-wide implementation and not actually used in the VLA parts of the codec. This makes adding new VLA implementation (e.g. Arm SVE) more expensive than it needs to be.

This PR doesn't add SVE support, but does some cleanup to minimize the vector library API as a precursor to doing so. The main changes are:

* Remove VLA indexable lane read functions, as with true VLA code the lane count isn't known.
* Replace VLA use of `.lane<0>()` with dedicated scalar function returns e.g. use `hmax_s()` rather than `hmax.lane<0>()`. This was beeing done in places before, but was not done consistently. Now this pattern is used everywhere.

### DIFF
--- a/Source/UnitTest/test_simd.cpp
+++ b/Source/UnitTest/test_simd.cpp
@@ -372,16 +372,6 @@ TEST(vfloat4, Loada)
 	EXPECT_EQ(a.lane<3>(), 3.0f);
 }
 
-/** @brief Test vfloat4 lane_id. */
-TEST(vfloat4, LaneID)
-{
-	vfloat4 a = vfloat4::lane_id();
-	EXPECT_EQ(a.lane<0>(), 0.0f);
-	EXPECT_EQ(a.lane<1>(), 1.0f);
-	EXPECT_EQ(a.lane<2>(), 2.0f);
-	EXPECT_EQ(a.lane<3>(), 3.0f);
-}
-
 /** @brief Test vfloat4 swz to float4. */
 TEST(vfloat4, swz4)
 {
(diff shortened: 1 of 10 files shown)


Source/astcenc_pick_best_endpoint_format.cpp
```diff
@@ -1307,7 +1307,7 @@ unsigned int compute_ideal_endpoint_formats(
 		vmask lanes_min_error = vbest_ep_error == hmin(vbest_ep_error);
 		vbest_error_index = select(vint(0x7FFFFFFF), vbest_error_index, lanes_min_error);
 		vbest_error_index = hmin(vbest_error_index);
-		int best_error_index = vbest_error_index.lane<0>();
+		int best_error_index = vbest_error_index.lane0();
```
bengaineyarm: You squished out the other lane0 accesses into hmax_s etc... any reason not to wrap this one?

Source/UnitTest/test_simd.cpp
```diff
@@ -169,8 +203,8 @@ TEST(vfloat, ChangeSign)
 /** @brief Test VLA atan. */
 TEST(vfloat, Atan)
 {
-	vfloat a(-0.15f, 0.0f, 0.9f, 2.1f);
-	vfloat r = atan(a);
+	vfloa4 a(-0.15f, 0.0f, 0.9f, 2.1f);
```
bengaineyarm: vfloa4

bengaineyarm: Nevermind you fixed it later

Source/astcenc_pick_best_endpoint_format.cpp
bengaineyarm: Nevermind you fixed it later

bengaineyarm (approved):
LGTM

PR #480, opened 2024-07-02
Add Arm SVE 8-wide (256b) implementation

solidpixel:
This PR adds support for Arm SVE via a fixed-width 256-bit implementation, as well as extending the fixed-width 128-bit implementation (which is mostly NEON) with a few targeted SVE operations such as native gathers. 

Due to the style of 128-bit accumulator we use for floating-point invariance, it's not really possible to write a true vector-length-agnostic SVE implementation, so this implementation is a compile-time choice that will only work on 256b SVE implementations.

On a Neoverse V1 this code is ~30% faster than the equivalent NEON build.


solidpixel:
This should check SVE width at runtime and error. As with other ISA checks that is done in the CLI front-end, and not in the library.

solidpixel:
SVE runtime checks now added.

Source/astcenc_vecmathlib.h
```diff
@@ -106,6 +114,33 @@
...
 	constexpr auto load1 = vfloat4::load1;
 
+#elif ASTCENC_SVE == 8
+	/* If we have SVE configured as 8-wide, expose 8-wide VLA. */
+	#include "astcenc_vecmathlib_neon_4.h"
+	#include "astcenc_vecmathlib_common_4.h"
+	#include "astcenc_vecmathlib_sve_8.h"
+
+	/* Check the compiler is treating SVE as 256 bits ... */
+	#if __ARM_FEATURE_SVE_BITS != 256
```
bengaineyarm, line 124: is there some possibility where this can be undefined by ASTCENC_SVE == 8?

perhaps better to have `#if !defined(__ARM_FEATURE_SVE_BITS) || (__ARM_FEATURE_SVE_BITS != 256)`

Source/astcenc_vecmathlib_neon_4.h
```diff
@@ -813,6 +805,11 @@ ASTCENC_SIMD_INLINE vfloat4 gatherf(const float* base, vint4 indices)
 	vals[2] = base[idx[2]];
 	vals[3] = base[idx[3]];
 	return vfloat4(vals);
+#else
+	svint32_t offsets = svset_neonq_s32(svundef_s32(), indices.m);
+	svfloat32_t	data = svld1_gather_s32index_f32(svptrue_pat_b32(SV_VL4), base, offsets);
```
bengaineyarm: odd spacing here

Source/astcenccli_entry.cpp
```diff
@@ -254,6 +255,36 @@ static bool validate_cpu_isa()
 	return true;
 }
 
+// Validate Arm SVE availability
+#elif ASTCENC_SVE != 0
+
+#include <sys/auxv.h>
+static bool cpu_supports_sve_256()
+{
+    long hwcaps = getauxval(AT_HWCAP);
```
bengaineyarm: Presumably checking the h/w vector length is not required here?

bengaineyarm: Actually it looks like you are checking the width later, so perhaps change the name?

Source/astcenc_vecmathlib_neon_4.h
solidpixel: Fixed in a later version.

Source/astcenc_vecmathlib.h
solidpixel: Good shout. Adding that would catch cases where the user has set the ASTCENC_SVE but fudged the compile flags and not setting a fixed-length.

solidpixel:
Confirmed passes all tests on V1 (as we can't test SVE using GH actions).

PR #492, opened 2024-08-09
Improve SIMD vtable API

### LINKED ISSUE #491: Improve SIMD table abstraction
The current SIMD code exposes tables as as fixed number of vector variables, but exactly how many variables you need for the table depends on the ISA implementation. These results in an untidy implementation where we pass around table variables that unused on some SIMD implementations, and I suspect the abstraction causes a sub-optimal SVE implementation.  

We should implement a set of new opaque `vfloat_tableN` data types, where the concrete type is defined by the SIMD library in use.

solidpixel:
This PR improves the API for the vtable functions, hiding the implementation behind an opaque struct that is defined by each 
SIMD implementation. 

No expected change to performance from the API change, this is just a maintainability improvement. 

Fixes #491

### DIFF
--- a/Source/astcenc_vecmathlib_avx2_8.h
+++ b/Source/astcenc_vecmathlib_avx2_8.h
@@ -1005,6 +1005,7 @@ ASTCENC_SIMD_INLINE void vtable_prepare(
 ) {
 	// AVX2 tables duplicate table entries in each 128-bit half-register
 	vint4 d0 = vint4::load(data);
+
 	table.t0 = vint8(astcenc_mm256_set_m128i(d0.m, d0.m));
 }
 
@@ -1016,14 +1017,14 @@ ASTCENC_SIMD_INLINE void vtable_prepare(
 	const uint8_t* data
 ) {
 	// AVX2 tables duplicate table entries in each 128-bit half-register
-	// Direct lookup for first row
 	vint4 d0 = vint4::load(data);
-	table.t0 = vint8(astcenc_mm256_set_m128i(d0.m, d0.m));
-
-	// XOR with previous rows for subsequent rows
 	vint4 d1 = vint4::load(data + 16);
-	d1 = d1 ^ d0;
+
+	table.t0 = vint8(astcenc_mm256_set_m128i(d0.m, d0.m));
 	table.t1 = vint8(astcenc_mm256_set_m128i(d1.m, d1.m));
+
+	// XOR chain the high rows to allow table emulation
+	table.t1 = table.t1 ^ table.t0;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


Source/astcenc_vecmathlib_avx2_8.h
```diff
@@ -971,98 +971,138 @@ ASTCENC_SIMD_INLINE vfloat8 int_as_float(vint8 a)
...
-{
-	// AVX2 duplicates the table within each 128-bit lane
-	__m128i t0n = t0.m;
-	t0p = vint8(astcenc_mm256_set_m128i(t0n, t0n));
+ASTCENC_SIMD_INLINE void vtable_prepare(
+	vtable8_32x8& table,
+	const uint8_t* data
+) {
+	// AVX2 tables duplicate table entries in each 128-bit half-register
+	// Direct lookup for first row
```
solidpixel, line 1019: Reorder this to match the layout below.

Source/astcenc_vecmathlib_neon_4.h
```diff
@@ -987,13 +1017,15 @@ ASTCENC_SIMD_INLINE vint4 vtable_8bt_32bi(vint4 t0, vint4 idx)
...
- * @brief Perform an 8-bit 32-entry table lookup, with 32-bit indexes.
+ * @brief Perform a vtable lookup in a 32x 8-bit table with 32-bit indices.
  */
-ASTCENC_SIMD_INLINE vint4 vtable_8bt_32bi(vint4 t0, vint4 t1, vint4 idx)
-{
+ASTCENC_SIMD_INLINE vint4 vtable_lookup(
+	const vtable4_32x8& tbl,
+	vint4 idx
+) {
 	int8x16x2_t table {
```
solidpixel, line 1026: Store the packed type directly in the table.

Source/astcenc_vecmathlib_sse_4.h
```diff
@@ -1037,63 +1037,92 @@ ASTCENC_SIMD_INLINE vfloat4 int_as_float(vint4 v)
...
 	__m128i idxx = _mm_or_si128(idx.m, _mm_set1_epi32(static_cast<int>(0xFFFFFF00)));
 
-	__m128i result = _mm_shuffle_epi8(t0.m, idxx);
+	__m128i result = _mm_shuffle_epi8(tbl.t0.m, idxx);
 	return vint4(result);
 #else
 	uint8_t table[16];
 
-	std::memcpy(table +  0, &t0.m, 4 * sizeof(int));
+	std::memcpy(table +  0, &tbl.t0.m, 4 * sizeof(int));
```
solidpixel, line 1125: Copy same implementation as None, and just store pointer in the table.

bengaineyarm (approved):

PR #494, opened 2024-08-10
Simplify SVE table lookup

solidpixel:
The design of the SVE vtbl allows us to simply bitwise OR successive lookups for a multi-lookup table because, unlike x86 shuffles, it's guaranteed that high values outside of the active 32-bit slice return zero. 

This has no impact on compression, but gives a small (0.5%) improvement to decompression performance.


Source/UnitTest/test_decode.cpp
```diff
@@ -65,7 +65,7 @@ TEST(decode, decode12x12)
 
 	status = astcenc_decompress_image(context, data, 16, &image, &swizzle, 0);
 	EXPECT_EQ(status, ASTCENC_SUCCESS);
-
+#if 0
```
solidpixel, line 68: Disable logging in the unit tests as it's just line noise.

Source/astcenc_vecmathlib_sve_8.h
```diff
@@ -910,22 +910,22 @@ ASTCENC_SIMD_INLINE vfloat8 int_as_float(vint8 a)
  * Table structure for a 16x 8-bit entry table.
  */
 struct vtable8_16x8 {
-	vint8 t0;
+	svuint8_8_t t0;
```
solidpixel, line 913: Use native types here to avoid casting on use.

bengaineyarm (approved):

bengaineyarm (approved):

PR #497, opened 2024-08-13
Assign names to threads in CLI tool

solidpixel:


Source/astcenccli_platform_dependents.cpp
```diff
@@ -165,6 +187,18 @@ double get_time()
 	return static_cast<double>(tv.tv_sec) + static_cast<double>(tv.tv_usec) * 1.0e-6;
 }
 
+/* See header for documentation */
+void set_thread_name(
+	const char* name
+) {
+	// No standard mechanism, so be defensive here
```
bengaineyarm, line 194: For linux at least you can use https://man7.org/linux/man-pages/man2/PR_SET_NAME.2const.html

    prctl(PR_SET_NAME, name, 0, 0, 0);

but it all amounts to the same thing i suppose

Source/astcenccli_platform_dependents.cpp
```diff
@@ -142,6 +146,24 @@ double get_time()
...
 }
 
+/* See header for documentation */
+void set_thread_name(
+	const char* name
+) {
+	// Names are limited to 16 characters
+	wchar_t wname [17] { 0 };
+	size_t name_len = std::strlen(name);
+	size_t clamp_len = std::min(name_len, static_cast<size_t>(16));
```
bengaineyarm, line 156: you could just do:

    std::min<size_t>(name_len, 16);

Source/astcenccli_toplevel.cpp
```diff
@@ -232,6 +232,10 @@ static void compression_workload_runner(
 ) {
 	(void)thread_count;
 
+	char name[17] { 0 };
+	std::snprintf(name, 16, "astc workc %d", thread_id);
```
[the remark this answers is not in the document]
solidpixel: Ah, I was trying to guarantee a `\0` terminator, but looks like snprintf already guarantees one, so yes.

PR #502, opened 2024-09-05
Move SVE length check to a veneer

solidpixel:
The current code tries to use an SVE length check to validate that the compile-time length matches the run-time length in the current hardware. However, it appears that the behavior of `svcntw()` is to return the compile-time fixed size if used in fixed-size code. 

This PR moves the validation check into a new veneer that is compiled with SVE, but without a fixed vector length.


Source/cmake_core.cmake
```diff
@@ -340,7 +352,7 @@ macro(astcenc_set_properties ASTCENC_TARGET_NAME ASTCENC_IS_VENEER)
                 ASTCENC_POPCNT=1
                 ASTCENC_F16C=0)
 
-        if (${ASTCENC_IS_VENEER})
+        if (NOT ${ASTCENC_VENEER_TYPE})
```
solidpixel, line 355: Inverted.

Source/cmake_core.cmake
```diff
@@ -365,7 +377,7 @@ macro(astcenc_set_properties ASTCENC_TARGET_NAME ASTCENC_IS_VENEER)
                 ASTCENC_POPCNT=1
                 ASTCENC_F16C=1)
 
-        if (${ASTCENC_IS_VENEER})
+        if (NOT ${ASTCENC_VENEER_TYPE})
```
solidpixel, line 380: Inverted.

markbarnett-arm (approved):

PR #503, opened 2024-09-05
Add support for fixed-width 128-bit SVE implementation

solidpixel:
This PR adds basic support for a 128-bit SVE implementation, which is implemented as an enhanced version of the NEON SIMD library rather than as a dedicated pure SVE implementation. 

Note that the NEON code was added as part of the 256-bit SVE port, so this change is purely build infrastructure to expose the new build type with the correct compiler settings.

On Neoverse V2 performance is 5% better for 4x4 blocks, improving to 10% better for 12x12 blocks. This is 1/3 from compiler-injected code, and 2/3 from using native gathers.


markbarnett-arm (changes requested):
`astcenc_print_header()` needs updating. It doesn't cater for `ASTCENC_SVE==4` so it'll end up printing simdtype "neon" in the help string.

markbarnett-arm (approved):

PR #504, opened 2024-09-21
Make sve_128 portable and support true march=native builds

solidpixel:
This PR makes two changes to improve support for SVE.

The first change makes the sve_128 build portable to any system implementing SVE by removing the fixed vector length at compile time. This code is merely augmented NEON, using native gathers, and will use predicates to force the 128-bit data path no matter what the host vector size is. Testing on Graviton 4 shows that there is no measurable performance difference between portable sve_128 and compile-time enforced sve_128, and testing on Graviton 3 shows that the portability part works fine.

The second change makes the "native" build type actually use `-march=native` and `-mcpu=native`, allowing native builds to pick up SVE if the host CPU supports it, as currently native builds rely on compiler defaults which are baseline Arm-v8. Note that this needs Clang-18 or newer to pick up SVE automatically, which is newer than the default compiler on current AWS Linux images for Graviton 3 and 4.

### DIFF
--- a/Source/astcenc_mathlib.h
+++ b/Source/astcenc_mathlib.h
@@ -74,7 +74,18 @@
 #endif
 
 #ifndef ASTCENC_SVE
-  #define ASTCENC_SVE 0
+  #if defined(__ARM_FEATURE_SVE)
+    #if defined(__ARM_FEATURE_SVE_BITS) && __ARM_FEATURE_SVE_BITS == 256
+      #define ASTCENC_SVE 8
+    // Auto-detected SVE can only assume vector width of 4 is available, but
+    // must also allow for hardware being longer and so all use of intrinsics
+    // must explicitly use predicate masks to limit to 4-wide.
+    #else
+      #define ASTCENC_SVE 4
+    #endif
+    #else
+    #define ASTCENC_SVE 0
+  #endif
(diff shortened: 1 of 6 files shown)


Source/astcenc_mathlib.h
```diff
@@ -74,7 +74,18 @@
 #endif
 
 #ifndef ASTCENC_SVE
-  #define ASTCENC_SVE 0
+  #if defined(__ARM_FEATURE_SVE)
+    #if defined(__ARM_FEATURE_SVE_BITS) && __ARM_FEATURE_SVE_BITS == 256
+      #define ASTCENC_SVE 8
+    // Auto-detected SVE can only assume vector width of 4 is available, but
+    // must also allow for hardware being longer and so all use of intrinsics
+    // must explicitly use predicate masks to limit to 4-wide.
```
bengaineyarm, line 82: I suppose we could (in future) do the same for >8wide? (with ASTENC_SVE 8)

bengaineyarm (approved):

Dreamer1:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #528, opened 2024-11-22
Using MRSSE for HDR error calculations

solidpixel:
*Status:*

Suspect I have the wrong end of the stick here - MRSEE on log values is still going to be incorrect. For MRSEE to make sense we need to use linear colors, diving in/out of LNS encoding when we need to test quantization. This is going to end up a lot more expensive ...

*Description*

HDR texture values are stored logarithmically.

Using absolute sum of squares on the logarithmic values (the existing behaviour on main) causes the compressor to spend too much effort preserving imperceptible shifts in dark channel values at the expense of bright values in the same block. This performs poorly in blocks with sharp luminance changes (dark texels in a bright block) and in blocks with saturated color values (dark channels in bright pixels).

Using absolute sum of squares on linearized HDR values (attempted by the first patch in this PR) avoids the compressor fixating on dark values, but instead causes the compressor to spend too much effort preserving bright values. This is because the errors in the bright channels can be orders of magnitude bigger than the errors in the dark channels, and dark values can end up quantizing close to black.

Using relative sum of square on the logarithmic values, proposed by Ryg in the blog below, encourages the compressor to find a balance of relative error across the whole block, favoring neither light nor dark channels. Blog: https://fgiesen.wordpress.com/2024/11/14/mrsse/ .

This is a perceptual error metric for HDR textures, so we expect that this will make the mPSNR score worse because we are no longer optimizing for absolute error. Test failures in HDR textures are therefore expected, and LDR textures are not affected.

- [ ] Suspect MRSSE needs to be in linear space, so try that ...
- [ ] Consider using MRSSE during iterative refinement. It's currently only used in final block check.
- [ ] If good, we need to regenerate test reference scores before merging so Actions keeps working.

### DIFF
--- a/Source/astcenc_compress_symbolic.cpp
+++ b/Source/astcenc_compress_symbolic.cpp
@@ -633,6 +633,7 @@ static float compress_symbolic_block_for_partition_1plane(
 
 				if (errorval < best_errorval_in_scb)
 				{
+					trace_add_data("select", "1");
 					best_errorval_in_scb = errorval;
 					workscb.errorval = errorval;
 					scb = workscb;
@@ -681,6 +682,7 @@ static float compress_symbolic_block_for_partition_1plane(
 
 			if (errorval < best_errorval_in_scb)
 			{
+				trace_add_data("select", "1");
 				best_errorval_in_scb = errorval;
 				workscb.errorval = errorval;
 				scb = workscb;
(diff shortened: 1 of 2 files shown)


solidpixel:
Rejecting for now - needs more work.

PR #548, opened 2025-02-06
Align with KDFS 1.4.0 rounding update

### LINKED ISSUE #547: Revert sRGB rounding mode changes
Khronos has now clarified the weight expansion and rounding behavior for ASTC in the latest Data Format Specification (not yet published, but text agreed). This update reverts an unintended behavior change which was introduced in the specification, which was the original trigger for this change:

* https://github.com/ARM-software/astc-encoder/pull/448

We need to revert this to align with the KDFS 1.4.0 release (now published). Khronos members discussion is tracked here for reference:

* https://gitlab.khronos.org/vulkan/vulkan/-/issues/4071

solidpixel:
Updates the codec rounding for sRGB to align with the KDFS 1.4.0 specification update, which reverts the alpha channel endpoint expansion behavior to match an earlier version of the specification. 

This changes the alpha channel to use the same method (`(x << 8) | 0x80`) as the RGB channels. The previous behavior had alpha doing (`(x << 8) | x`). The impact of this change is differences in the LSB.

Fixes #547


Dreamer1:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #554, opened 2025-03-09
simplify mask() implementation in none_4

camel-cdr:
Just a small thing I've noticed.
Using the -1/0 values for true/false allows this easier conversion to a packed representation.


solidpixel (approved):

solidpixel:
Thanks - merged.

PR #566, opened 2025-03-15
Use latest GTest stable tag

### LINKED ISSUE #565: Update GTest to reduce build spam
We're currently using an older version of GTest, which doesn't use CMake in the way modern CMake likes, so it's triggering a lot of Deprecation warnings in the builds on configure. 

Updating GTest should fix this one.

solidpixel:
Fixes #565


solidpixel:
Work in progress - looks like the new version of GTest + MSVC turn "or" and "and" used as test names in the TEST() macro into operators, so we need to avoid those and use something less generic for the test names.

PR #583, opened 2025-09-04
Add KTX orientation metadata

bmdhacks:
I'm not sure if this is correct, but my read is that most software expects ktx to be using DirectX orientation, and reading the KTX file directly into OpenGL will result in inverted textures, so we've been adding yflip but there's nothing to indicate this in the metadata.


bmdhacks:
Sorry I used my main branch to do a bunch of other changes and fucked up this PR, I'll resubmit.

PR #584, opened 2025-11-02
Fixed astcenc.h for C builds and double definition of NOMINMAX

m256i:
Firstly, i could not determine if full C compatibility was intended for this repository, so i want to apologize if it was not.

## issues

- `<astcenc.h>`  unconditionally includes `<cstddef>` and `<cstdint>` which are C++ exclusive headers.
- the struct and union definitions in `<astcenc.h>` are not typedef'd to global scope which makes most struct and function definitions ill-formed in a C context
- `extern "C"` is used unconditionally in `<astcenc.h>` with `ASTCENC_DYNAMIC_LIBRARY` which is also invalid C
- NOMINMAX is defined twice under my build system (`ninja 1.11.1` + `MinGW UCRT64 gcc 15.1.0`)

These changes are ABI BREAKING since C++ name mangling/linkage for all API exposed functions in `<astcenc.h>` get disabled with this commit. 

## changes

- wrapped c++ only includes `<cstddef>` and `<cstdint>` around `#if defined(__cplusplus)`, use C compat versions otherwise
- additionally included `<stdbool.h>` for C builds
- added ASTCENC_EXTERN_C macro that evaluates to `extern "c"` only for C++ builds
- wrapped `#define NOMINMAX` into a guard that prevents double definitions

Perhaps i should add a "ASTCENC_C_COMPAT" flag to the CMake build file to avoid a forced ABI change for all users?

I have tested building the library with these changes on both C and C++ compilers with success. I am open to any other approach that the maintainers prefer, if C compatibility IS intended :)

### DIFF
--- a/Docs/ChangeLog-5x.md
+++ b/Docs/ChangeLog-5x.md
@@ -18,6 +18,8 @@ header.  We always recommend rebuilding your client-side code using the
 header from the same release to avoid compatibility issues.
 
 * **General:**
+  * **Improvement:** The interface header `astcenc.h` is now C compliant to
+    make it usable from C programs.
   * **Improvement:** Contexts using the same configuration can now share
     read-only data tables. This can significantly reduce the amount of memory
     needed for applications that parallelize by processing multiple images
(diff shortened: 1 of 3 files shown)


solidpixel:
Full C compatability wasn't something I originally aimed for - the `extern "C"` declarations are there so there is a stable compiler-independent name for resolving at run-time though `dlsym()`. 

I'm very happy to support it if you need it though, it's not going to cost anything in terms of complexity or performance.

solidpixel:
Rebased on latest.

solidpixel:
Thanks for the PR and sorry for the very slow turn around time on getting it merged!

PR #593, opened 2026-03-31
Fix clang20+ compilation

KaperD:
Since clang 20 there is overriding-option check, which fails if both -ffp-model=precise and -ffp-contract=off are set (they are for AppleClang e.g.). Example https://godbolt.org/z/eWMn4c7z9

From clang docs: `precise Disables optimizations that are not value-safe on floating-point data, although FP contraction (FMA) is enabled (-ffp-contract=on). This is the default behavior. This value resets -fmath-errno to its target-dependent default.`

So, `-ffp-model=precise` = `-fno-unsafe-math-optimizations -ffp-contract=on` + reset -fmath-errno. But you don't use errno, that is why we can change `-ffp-model=precise` to -fno-unsafe-math-optimizations`


solidpixel:
Thanks for the heads up and the PR - will take a look.

PR #594, opened 2026-04-15
add RVV VLEN=256 VLS backend

camel-cdr:
Hi, this PR adds an initial VLEN=256 RVV backend.

It passes all tests and gave a 2x speedup on the [SpacemiT X60](https://camel-cdr.github.io/rvv-bench-results/spacemit_x60/index.html) core.

```
# compiled with -march=rv64gcbv_zvl256b -mrvv-vector-bits=zvl
$ astc-encoder-rvv -tl Test/Images/Khronos/LDR-RGB/ldr-rgb-metalrough2.png 8x6 -fast -repeats 3 -j 1
Quality metrics
===============
    PSNR (LDR-RGBA):     35.0056 dB
    Alpha-weighted PSNR: 35.0056 dB
    PSNR (LDR-RGB):      33.7563 dB
Performance metrics
===================
    Total time:          26.0600 s
    Coding time:         20.8997 s
    Coding rate:          0.2007 MT/s
    Decoding time:        0.8910 s
    Decoding rate:        4.7076 MT/s
# compiled with -march=rv64gcbv_zvl256b, so with potential autovec of none_4 backend
$ astc-encoder-original -tl Test/Images/Khronos/LDR-RGB/ldr-rgb-metalrough2.png 8x6 -fast -repeats 3 -j 1
Quality metrics
===============
    PSNR (LDR-RGBA):     35.0056 dB
    Alpha-weighted PSNR: 35.0056 dB
    PSNR (LDR-RGB):      33.7563 dB
Performance metrics
===================
    Total time:          46.0870 s
    Coding time:         41.2464 s
    Coding rate:          0.1017 MT/s
    Decoding time:        1.0705 s
    Decoding rate:        3.9181 MT/s
```

You can find documentation for RVV here:
* [spec docs.riscv.org](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html)
* [unofficial intrinsics viewer](https://dzaima.github.io/intrinsics-viewer/)

The CI doesn't include a gcc build, because the upstream astc-encode doesn't pass the tests when compiling for RISC-V with RVV enabled.
This is presumably a gcc codegen bug, but may also be some implementation-defined behavior or UB in the astc code.
It's also quite slow (still running on my end 30min in, it's more resonable local with newer qemu), so you may want to disable it.

Because the current implementation relies on the `riscv_rvv_vector_bits` attribute, the RVV backend is only enabled, when astc is compiled with `-mrvv-vector-bits=zvl` and a `-march=.._zvl<N>b` matching the native vector length.
The code itself would work for any fixed-vector-length, but the default constant ASTC parameters aren't compatible with more than 8 elements in a vector and using it for 4 elements runs into conflicts with the none_4 backend.
I'm planning to refactor the generic >4-element vector abstraction next, to support larger vector length and VLA RISC-V and SVE codegen.

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -231,6 +231,62 @@ jobs:
         run: ctest --rerun-failed --output-on-failure
         working-directory: build_rel
 
+  build-ubuntu-rvv256-clang:
+    name: Ubuntu rvv256 Clang
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: 'true'
+
+      - name: Update apt packages
+        run: sudo apt-get update
+
+      - name: Install ImageMagick
+        run: sudo apt-get install imagemagick
+
+      - name: Install Python
+        run: sudo apt-get install python3 python-is-python3 python3-numpy python3-pillow
+
+      - name: Install clang-20 and RISC-V binutils
+        run: sudo apt-get install clang-20 g++-riscv64-linux-gnu
+
+      - name: Install RISC-V qemu
+        run: sudo apt-get install qemu-user-static
+
+      - name: Build release
+        run: |
+          export CXX=clang++-20
+          export CXXFLAGS="--target=riscv64-linux-gnu -march=rv64gcv_zvl256b -mrvv-vector-bits=zvl"
+          mkdir build_rel
+          cd build_rel
+          cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../ -DASTCENC_UNITTEST=ON -DASTCENC_ISA_NONE=ON ..
+          make install -j$(nproc)
+
+      - name: Build debug
+        run: |
+          export CXX=clang++-20
+          export CXXFLAGS="--target=riscv64-linux-gnu -march=rv64gcv_zvl256b -mrvv-vector-bits=zvl"
+          mkdir build_dbg
+          cd build_dbg
+          cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DASTCENC_ISA_NONE=ON ..
+          make -j$(nproc)
+
+      - name: Run system tests
+        run: |
+          export QEMU_CPU="max,vlen=256,rvv_ta_all_1s=on,rvv_ma_all_1s=on"
+          export QEMU_LD_PREFIX="/usr/riscv64-linux-gnu"
+          python ./Test/astc_test_functional.py --encoder none
+          python ./Test/astc_test_image.py --encoder none --test-set Small -j $(nproc)
+
+      - name: Run unit tests
+        run: |
+          export QEMU_CPU="max,vlen=256,rvv_ta_all_1s=on,rvv_ma_all_1s=on"
+          export QEMU_LD_PREFIX="/usr/riscv64-linux-gnu"
+          ctest --rerun-failed --output-on-failure
+        working-directory: build_rel
+
   build-macos-x64-clang:
     name: macOS x64 Clang
     runs-on: macos-14
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -231,62 +231,6 @@ jobs:
         run: ctest --rerun-failed --output-on-failure
         working-directory: build_rel
 
-  build-ubuntu-rvv256-clang:
-    name: Ubuntu rvv256 Clang
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Git checkout
-        uses: actions/checkout@v4
-        with:
-          submodules: 'true'
-
-      - name: Update apt packages
-        run: sudo apt-get update
-
-      - name: Install ImageMagick
-        run: sudo apt-get install imagemagick
-
-      - name: Install Python
-        run: sudo apt-get install python3 python-is-python3 python3-numpy python3-pillow
-
-      - name: Install clang-20 and RISC-V binutils
-        run: sudo apt-get install clang-20 g++-riscv64-linux-gnu
-
-      - name: Install RISC-V qemu
-        run: sudo apt-get install qemu-user-static
-
-      - name: Build release
-        run: |
-          export CXX=clang++-20
-          export CXXFLAGS="--target=riscv64-linux-gnu -march=rv64gcv_zvl256b -mrvv-vector-bits=zvl"
-          mkdir build_rel
-          cd build_rel
-          cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../ -DASTCENC_UNITTEST=ON -DASTCENC_ISA_NONE=ON ..
-          make install -j$(nproc)
-
-      - name: Build debug
-        run: |
-          export CXX=clang++-20
-          export CXXFLAGS="--target=riscv64-linux-gnu -march=rv64gcv_zvl256b -mrvv-vector-bits=zvl"
-          mkdir build_dbg
-          cd build_dbg
-          cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DASTCENC_ISA_NONE=ON ..
-          make -j$(nproc)
-
-      - name: Run system tests
-        run: |
-          export QEMU_CPU="max,vlen=256,rvv_ta_all_1s=on,rvv_ma_all_1s=on"
-          export QEMU_LD_PREFIX="/usr/riscv64-linux-gnu"
-          python ./Test/astc_test_functional.py --encoder none
-          python ./Test/astc_test_image.py --encoder none --test-set Small -j $(nproc)
-
-      - name: Run unit tests
-        run: |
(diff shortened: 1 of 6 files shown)


solidpixel:
Are you willing to be considered the maintainer for this one going forward? I don't mind accepting the PR, but I don't really want to be maintaining another backend if there are any issues with it in future. 

> I'm planning to refactor the generic >4-element vector abstraction next, to support larger vector length and VLA RISC-V and SVE codegen.

The main challenge with true VLA code gen is that we have a strict requirement for floating point invariance across platforms (the major game engines need this to get cross-platform consistency). Performant VLA code changes floating point reassociation in loop accumulators, and therefore is not invariant across platforms. 

See https://solidpixel.github.io/creating_invariant_accumulators which I wrote when adding AVX2.

I tried it for SVE in the past and to keep invariance you need to make the accumulator look like the 4-wide accumulator we use on all the other instruction sets, which means you need a loop to iterate over vector length in chunks of 4. The compressor needs a lot of accumulators on the hot path, and adding accumulator loops to _every_ hot path in the code is rather expensive.

camel-cdr:
> Are you willing to be considered the maintainer for this one going forward?

Yes, I can do that.

> The main challenge with true VLA code gen is that we have a strict requirement for floating point invariance across platforms

I was planning to do reduction loops that maintain the invariance, like the curremt SVE backend, but iteration count depending on the runtime VLEN. BTW, since that is a requirement, the simd unittest should really use EXPECT_EQ instead of EXPECT_NEAR, when the invariance isn't disabled.

> The compressor needs a lot of accumulators on the hot path, and adding accumulator loops to every hot path in the code is rather expensive.

I can see this beeing expensive for current Arm cores, which have the same execution resources available for NEON and SVE.
Since RISC-V doesn't have a fixed width SIMD ISA, cores actually implement a >128-bit execution units, so accumulation loop would have to be faster than a vslidedown+vfadd for it not to be worth it.

I suppose compute_parition_averages_rgb could be problematic, but it's probably best to benchmark.

The invariance requirement is relative to previous versions, not only between targets, right? Because there are some other ways to do invariant accumulation, but the invariance would different from the current one.

solidpixel:
> BTW, since that is a requirement, the simd unittest should really use EXPECT_EQ instead of EXPECT_NEAR, when the invariance isn't disabled.

Yes, agreed. It does in most places - looks like I didn't when I added the atan2 tests. Will fix.

> The invariance requirement is relative to previous versions, not only between targets, right? Because there are some other ways to do invariant accumulation, but the invariance would different from the current one.

Requirement is for invariance across all targets for a specific version of the compressor - we can change output across versions (compressor heuristic changes will change output anyway). 

If we can find a performance neutral way of doing it that supports VLA and does not hurt the other targets I'd definitely be willing to consider it. It's less interesting if there is a performance regression on other targets given that they are the variants that actually get used by developers today.

camel-cdr:
> Requirement is for invariance across all targets for a specific version of the compressor - we can change output across versions (compressor heuristic changes will change output anyway).

Very cool.

There are variants that also speedup the 4-element implementations. The simplest is just using more accumulators to change latency to throughput bound, but that's challanging when there are many accumulators.

Even full tree reduction is faster than a single 4-element accumuator, see how the autovectorized 4linear is slower than tree SIMD: https://godbolt.org/z/48E87xebs

But I suspect the best way would be to use wider fixed accumulators and unroll for vec4, when that's possible without spilling 16 registers (sse). That will mostly be a question of how to design an API that isn't to ugly.

---

I'd also like to add a [gcc/clang vector extension](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html) vec4 implementation, mostly for RISC-V, but it would be generic. The problem is, you can't implement a vector length agnostic fixed width SIMD abstraction (that are stored in structs) with the RVV intrinsics, because the types are sizeless, so can't be stored in a struct, and compilers don't support redundant predicated load/store elimination. The gcc vector extension does however work for this usecase and has perfect portable codegen for RVV, when you can express the problem with it. (It just always sets the vector length to 4)
A quick test implementation reduced a hot function from 380 (scalar backend with autovec) to 180 instructions.

camel-cdr:
Scratch the gcc vector extension idea.
Turns out that isn't stabel enough and gcc runs into an internal compiler error in gcc when compiling it for RISC-V.
It also wasn't faster than the none backend in gcc when compiling for x86, although it almost matched the SSE2 backend in clang:
```
gcc   SSE2: 2.6848 s
gcc   none: 4.9846 s
gcc   vext: 7.7711 s
clang vext: 2.8707 s
```
Still on my RISC-V board it only improved the benchmark from 26 to 24 seconds, which isn't worth the hassle.

Here is the implementation: https://godbolt.org/z/hYvT4Mva8

solidpixel:
@camel-cdr Some general review questions:

In the new RVV SIMD backend, the table functions have multiple implementations gated by `#if` on the vector length. What testing is needed to ensure all of those variants get some test coverage (if only through the SIMD unit tests)? Can we cover all of those through QEMU in the GitHub Actions test list?

We don't require copyright assignment, so feel free to add yourself as a new line in the copyright list in the files you modified (you don't have to - but feel free to do so if you wish).

Can you add some notes on RISC-V vector length compatibility and build options required to the "Advanced build options" section of `docs/Building.md`. 

I also have some minor code style cleanups in a local branch. I can apply it later, but if possible can you allow me to commit to your PR branch so we can do it all in one go?

solidpixel:
Created #596 to track the fixes to unit tests to use EXPECT_EQ vs EXPECT_NEAR

camel-cdr:
@solidpixel qemu supports VLEN=128/256/512/1024 and the spike simulator all possible VLEN (with a single line patch).
I only enabled the backend for VLEN=256, but I've kept the implementations generic, which required specialization for VLEN=128, VLEN=256 and VLEN>=512 for good performance.
Since the default astc config doesn't allow for VLEN>256 (there are checks of SIMD_WIDTH, that only work up to 256-bit), I wasn't able to test the VLEN=512 codepath.
I did however test that the vfloat4/vint4 unit tests succeed, when I slightly modify the code to use the RVV backend for vfloat4/vint4.

> We don't require copyright assignment, so feel free to add yourself as a new line in the copyright list in the files you modified (you don't have to - but feel free to do so if you wish).
> Can you add some notes on RISC-V vector length compatibility and build options required to the "Advanced build options" section of docs/Building.md.

:+1:

> I also have some minor code style cleanups in a local branch. I can apply it later, but if possible can you allow me to commit to your PR branch so we can do it all in one go?

I've added you as a collaborator to my fork. If that is what you meant?

solidpixel:
> I've added you as a collaborator to my fork. If that is what you meant?

That works if you're happy with that. There used to be a tick box for a PR that only allowed maintainers to touch the branch used for the PR: 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

camel-cdr:
@solidpixel ah, that was already enabled.

solidpixel:
> @solidpixel ah, that was already enabled.

Hmm, weird, I was getting an error when I tried earlier. 

I've pushed the stylistic cleanup patch to the branch - nothing functional, just whitespace and moving a few things to align with the coding standard used elsewhere in the repo.

camel-cdr:
@solidpixel I think you mistyped the branch name. You pushed to `rvv_256_vl`, while the PR branch is `rvv_256_vls`.

solidpixel:
Argh, sorry. Try that.

solidpixel:
> Since the default astc config doesn't allow for VLEN>256 (there are checks of SIMD_WIDTH, that only work up to 256-bit), I wasn't able to test the VLEN=512 codepath.

Right, makes sense. 

If we can't test it, I think my preference is to add a comment to those code blocks explaining that it's not reachable and is untested, and add a compile-time assert that will fault the build if you try to compile for it. Feels less risky than leaving untested code there as a future pit-trap.

Happy to leave the wider code implementation in place, as long as the lack of testing is commented in each instance.

solidpixel:
For the RISC-V testing being slow problem, we should only need to run RISC-V testing specifically if the RVV SIMD code changes. The SIMD library interface is robust and well tested, so I'm willing to relax RV test coverage to just changes to the SIMD library implementation. 

If we split RISC-V it into its own Actions YAML file, with something like:

```yaml
on:
  workflow_dispatch:
  pull_request:
    branches:
      - main
    paths:
      - 'Source/astcenc_vecmathlib_rvv_n.h'
```

... to control when it's run, then most test runs won't slow down, and we avoid disabling RV testing completely.

camel-cdr:
@solidpixel can you commit that? I'd probably need a few tries to get it right.

solidpixel:
Yep, will do.

solidpixel:
OK, that looks good for now. I'll let the tests run and take a look in the morning, and sign off the review if all is looking green. Thanks for the quick iteration.

camel-cdr:
BTW, it may be worth adding `-fno-math-errno` to the build options for better autovectorization of sqrt in the none backend: https://godbolt.org/z/6EjevsYKa
As far as I could tell the code doesn't read errno, so it should be fine. Should be independent of this PR though.

solidpixel:
Yeah, that would make sense to do - I've generally not worried too much about the NONE backend performance historically.

solidpixel:
Raised #598 for errno changes

solidpixel (approved):
Reviewed - looks good, thanks!

PR #621, opened 2026-05-30
use size_t for decompress size check to avoid overflow

sahvx655-wq:
Decoding an .astc whose header gives a large texel size (dims and block size come straight from the file in load_cimage) lets xblocks*yblocks*zblocks*16 wrap in 32-bit before it widens to size_t. A 262144x16384 image at 4x4 is 2^28 blocks, so the byte count is 2^32 and the computed size_needed truncates to 0. The data_len < size_needed guard then passes for a 16-byte payload and the decode loop walks data + i*16 well past the buffer. Doing the multiply in size_t keeps the check honest.

### DIFF
--- a/Source/astcenc_entry.cpp
+++ b/Source/astcenc_entry.cpp
@@ -1215,7 +1215,7 @@ astcenc_error astcenc_decompress_image(
 	int plane_blocks = xblocks * yblocks;
 
 	// Check we have enough output space (16 bytes per block)
-	size_t size_needed = xblocks * yblocks * zblocks * 16;
+	size_t size_needed = static_cast<size_t>(xblocks) * yblocks * zblocks * 16;
 	if (data_len < size_needed)
 	{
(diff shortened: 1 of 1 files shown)


solidpixel:
Thanks for the PR. 

I agree this needs fixing, but there are other similar issues around this area that need fixing for robust error checking (e.g. dimensions can overflow as part of the rounding to block size check, some mixed int and unsigned int usage). I'll reject this PR and fix all of these in a larger change.

sahvx655-wq:
Makes sense, a single sweep over the dimension handling is cleaner than patching one multiply. The rounding-to-block-size step is the other spot I'd watch since the rounded dims can overflow before they even reach this check. Happy to test against the crafted .astc once your change lands.

solidpixel:
#623 makes library interface handle and detect overflows
#624 adds invalid size checking for .astc files in the CLI wrapper (merge pending - GitHub Actions having a bad day)

PR #626, opened 2026-06-02
use size_t for dds and ktx surface size to avoid overflow

sahvx655-wq:
Fuzzing the image loaders with crafted KTX and DDS headers turned this up. The surface size in load_ktx_uncompressed_image and load_dds_uncompressed_image is computed in uint32_t, so large width/height/depth truncate the byte count and the buffer is allocated too small while the scanline copy still walks the full dimensions, reading off the end of the heap. The KTX size check is bypassed by setting the stored surface size to the truncated value. Same shape as the recent .astc fix, so compute in size_t and reject on overflow here too.

### DIFF
--- a/Source/astcenccli_image.cpp
+++ b/Source/astcenccli_image.cpp
@@ -228,8 +228,6 @@ astcenc_image* astc_img_from_unorm8x4_array(
 	return img;
 }
 
-// initialize a flattened array of float values from an ASTC codec image
-// The returned array is allocated with new[] and must be deleted with delete[].
 /* See header for documentation. */
 float* floatx4_array_from_astc_img(
 	const astcenc_image* img,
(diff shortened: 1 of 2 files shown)


solidpixel:
Thanks for the PR - will review in the next few days.

solidpixel:
I've refactored the overflow checks into common library functions in #628, which should simplify the added code here. Will rebase this after the other PR has merged.

solidpixel:
Existing issues spotted while reviewing, I'll pick up in follow-on PRs for these:

* Error messages for KTX and DDS image loading do not use `print_error()`, and should be cleaned up to follow the standard message formatting too. Raised issue #629.
* Not all uses of `new[]` for file inputs are wrapped in "bad_alloc" catches - we should probably hoist a single check for those somewhere in common code higher up the call stack. Raised issue #630. 
* It would be good to use scoped memory/resource management in the frontend to avoid all of the manual cleanup in the error paths. This has been a long-time TODO, but we've never done it. We should because the current code makes ASAN throw leak errors currently if the wrapper catches an error and exits without releasing the resources, which means we can miss real leaks we care about. Raised issue #631.

PR #635, opened 2026-06-03
use size_t for image plane allocation in alloc_image

sahvx655-wq:
alloc_image sizes each plane with dim_x * dim_y * 4 evaluated as unsigned int, but those dimensions come straight from a .astc or .ktx header, so a 32768x32768 image wraps the multiply and allocates a 0 byte plane. The decompressor then writes the full 4*dim_x*dim_y extent with size_t indexing in store_image_block and runs off the end of the buffer. I hit this reading the alloc path after the recent surface-size fixes: the KTX and DDS loaders check the size in 64-bit but the allocation here was still 32-bit, and the decompress output path has no guard at all. Widen the multiply to size_t so the buffer matches what gets written.

### DIFF
--- a/Source/astcenccli_image.cpp
+++ b/Source/astcenccli_image.cpp
@@ -49,7 +49,7 @@ astcenc_image_ptr alloc_image(
 		img->data_type = ASTCENC_TYPE_U8;
 		for (unsigned int z = 0; z < dim_z; z++)
 		{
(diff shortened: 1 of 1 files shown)


solidpixel:
Good spot, thank you.

solidpixel (approved):

